PR #4, opened 2026-05-22
Add dynamic OpenCode metadata and Zen GPT/Gemini routing

Wallacy:
## Summary

This PR improves the OpenCode Copilot Chat extension in three areas:

- hardens OpenCode requests with configurable request/stream timeouts, sticky session headers, and clearer quota/rate-limit errors
- replaces static picker metadata with a TTL-cached `models.dev` snapshot merged with live `/models` metadata and bundled fallback values
- adds internal family-based routing so Zen GPT models use `/responses`, Zen Gemini models use the documented Google-style `generateContent` endpoint, Zen Claude models use `/messages`, and Go MiniMax models use `/messages`

It also updates the extension docs and prepares the `0.1.6` release.

## Why this approach

`models.dev` is a good source for limits, deprecation state, reasoning capability, and vision support, but it does not expose the transport shape a model expects.

Because of that, this PR keeps transport routing inside the extension instead of scraping docs at runtime. The internal route map is smaller, cheaper, and more predictable, while still matching the currently documented OpenCode families.

## What changed

- Added `opencodego.requestTimeoutSeconds` and `opencodego.streamIdleTimeoutSeconds`
- Added sticky OpenCode request headers:
  - `x-opencode-session`
  - `x-opencode-request`
  - `x-opencode-client`
- Improved rate-limit/quota errors surfaced in VS Code
- Added TTL-cached `models.dev` metadata in VS Code global state
- Merged live `/models` metadata, cached `models.dev` metadata, and bundled fallback metadata before registering models
- Updated fallback limits/capability hints used when live metadata is unavailable
- Added native Zen GPT support through `/responses`
- Added native Zen Gemini support through `/models/{model}:streamGenerateContent?alt=sse`
- Updated README and changelog for `0.1.6`

## Testing

- `npm run compile`
- `npx vsce package -o opencode-copilot-chat-0.1.6-local.vsix`
- Installed the local VSIX in VS Code with:
  - `code --install-extension .\opencode-copilot-chat-0.1.6-local.vsix --force`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 - Added a TTL-cached models.dev metadata snapshot, merged with live `/models` metadata and a bundled fallback catalog for offline picker registration.
 - Added native Zen GPT routing through `/responses` and Zen Gemini routing through the documented Google-style `/models/{model}:streamGenerateContent?alt=sse` endpoint.
 
+### Changed
+
+- Corrected fallback-advertised model limits to follow `models.dev` whenever the live `/models` payload does not provide limit metadata, fixing earlier Go/Zen cross-provider mix-ups in the bundled table.
+- This reduces several previously overstated fallback values, notably `deepseek-v4-flash-free` to `200000 / 128000`, `glm-5` and `glm-5.1` to `202752 / 32768`, and Go `minimax-m2.5` to `204800 / 65536`.
+
 ### Fixed
 
 - Updated bundled fallback limits and capability hints so the picker stays usable when neither `/models` nor models.dev can be refreshed.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ## Unreleased
 
+## [0.1.6] — 2026-05-21
+
+### Added
+
+- Added configurable OpenCode request and streaming idle timeouts so Copilot Chat requests fail clearly instead of hanging indefinitely.
+- Added sticky OpenCode request headers (`x-opencode-session`, `x-opencode-request`, `x-opencode-client`) so Go and Zen requests preserve gateway affinity behavior.
+- Added clearer rate-limit and quota handling, including retry/quota details from response headers when available.
+- Added a TTL-cached models.dev metadata snapshot, merged with live `/models` metadata and a bundled fallback catalog for offline picker registration.
+- Added native Zen GPT routing through `/responses` and Zen Gemini routing through the documented Google-style `/models/{model}:streamGenerateContent?alt=sse` endpoint.
+
+### Fixed
+
+- Updated bundled fallback limits and capability hints so the picker stays usable when neither `/models` nor models.dev can be refreshed.
+- Zen Claude, Zen GPT, Zen Gemini, and Go MiniMax families now use the correct transport automatically instead of being forced through a single OpenAI-compatible route.
+
 ## [0.1.4] — 2026-05-17
 
 ### Added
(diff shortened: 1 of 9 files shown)


ltmoerdani:
Hey @Wallacy thanks for the PR!

Compile clean, installed locally, routing is solid. A few things before merge.

The model limit changes need some explanation. Several dropped significantly:

| Model | Before | After |
|---|---|---|
| `deepseek-v4-flash-free` | ctx 1M, out 384K | ctx 200K, out 128K |
| `glm-5.1` / `glm-5` | ctx 204K, out 131K | ctx 202K, out 32K |
| `minimax-m2.5` (Go) | out 131K | out 65K |

Are these from the live `/models` endpoint or `models.dev`? If from `models.dev`, we should add a `### Changed` section to the CHANGELOG. Users relying on those higher limits will notice and wonder why things broke.

`refreshOpenCodeModelMetadata()` calls `fetch(MODELS_DEV_API_URL)` without a timeout. If `models.dev` goes slow or down, model registration just hangs until TCP gives up. One-liner fix:

```typescript
const response = await fetch(MODELS_DEV_API_URL, {
  signal: AbortSignal.timeout(10_000)
});
` ` `

The file went from ~1,960 to ~3,527 lines in a single PR. That's 80% growth, and it's going to get painful to review over time. Breaking it up makes sense:

- `routing.ts` — `resolveModelRouting()`, `normalize*StreamEvent()` parsers
- `metadata.ts` — models.dev cache, `resolveModelMetadata()`, fallback logic
- `streaming.ts` — `streamOpenCodeResponse()`, `streamChatCompletions()`, `streamResponsesApi()`, `streamGoogleGenerateContent()`
- `errors.ts` — `OpenCodeRequestError`, rate-limit parsing, `buildOpenCodeRequestError()`
```

Can be a follow-up if you want to keep this PR scoped.

The three new normalizers (`normalizeResponsesStreamEvent`, `normalizeGoogleStreamEvent`, `normalizeGoogleFullResponse`) are handling external format parsing — exactly where silent breakage tends to hide when providers change things upstream. Snapshot tests with mocked SSE data would give a safety net. Worth covering at minimum:

- Responses API: `response.output_text.delta`, `response.function_call_arguments.delta`, `response.completed`, unknown event type
- Google: candidate with `text` + `thought`, `functionCall` parts, `STOP` / `MAX_TOKENS` finish reason
- Routing: `gpt-*` → `responses`, `gemini-*` → `google`, `claude-*` → `messages`, `minimax-m2.*` → `messages`, everything else → `chat-completions`

Happy to help test once these are in.

Wallacy:
Thanks for the careful review. I pushed a follow-up that addresses the concrete points you raised.

On the limit changes: the reduced values came from models.dev, not the live /models payload. In practice the affected /models responses do not reliably expose those limit fields, so the extension now uses the 6-hour TTL-cached models.dev snapshot as the authoritative source whenever live metadata is missing. The earlier bundled table had also mixed some Go and Zen values, which is why a few fallback defaults were overstated. I added a Changed section to the changelog to call out the notable reductions.

I also added a 10s timeout to the models.dev refresh path so model registration does not sit on a slow network call indefinitely.

For structure, I split the pure helper logic out of extension.ts into separate metadata, routing, and error modules, and added parser/routing coverage for the new Responses and Google normalizers plus the family-based endpoint routing cases you listed.

Thanks again. This should make the risky parsing and routing paths easier to review and safer to maintain. If you still want, I can do a smaller follow-up to extract the streaming layer as well.

ltmoerdani:
Looks great, thanks for addressing all the points! The limit correction makes sense, if the previous values were mixing Go/Zen then this is a bug fix, not a regression.

Merging now. And yeah, a follow-up for the streaming layer extraction would be awesome whenever you have time 👍

PR #7, opened 2026-06-02
feat: add support for VS Code's languageModelPricing API

Wallacy:
## What

This PR surfaces real per-model cost data from the `models.dev` registry in
the VS Code model picker and management UI and tidies up the experimental context-indicator
config that became redundant after commit `ca8bbb6`.

## Why

VS Code recently added a `vscode.proposed.languageModelPricing` API that
exposes `pricing`, `inputCost`, `outputCost`, `cacheCost`, and
`priceCategory` on every registered language model. The OpenCode Go/Zen
provider was not populating those fields. The `models.dev` registry already publishes the underlying cost
data — we just weren't reading it.

## Changes

- Read `cost.input`, `cost.output`, `cost.cache_read`, and `cost.cache_write`
  from the `models.dev` snapshot in `src/metadata.ts`, including the
  long-context tiers when present, and convert USD to AI Credits
  (1 credit = $0.01 USD) so VS Code renders it alongside the official
  Copilot models.
- Surface those fields on every registered model via the proposed pricing
  interface added in `src/vscode.proposed.chatProvider.d.ts`. The
  `priceCategory` thresholds use Copilot's 3:1 input:output weighted blend so
  the buckets line up: Kimi k2.6 → `medium`, GPT-5.4 → `high`, Claude Opus
  4.5 → `high`, GPT-5.4 Pro → `very_high`.
- Bump the cached `models.dev` snapshot key from `v3` to `v4` so users
  automatically re-fetch the registry on next activation and pick up the
  newly parsed `cost` and modality fields instead of consuming stale cached
  entries that did not carry them.
- Detect input modalities (`image`, `audio`, `video`, `pdf`) from the
  `models.dev` `modalities.input` array and surface them in the picker
  tooltip / detail line alongside the existing vision indicator.
- Drop the `opencodego.experimentalContextIndicator` configuration and the
  associated hook bridge in `src/contextWindowHook.ts`. The same capability
  is already wired up natively after commit `ca8bbb6`, so the experimental
  path was redundant.
- Remove the local duplicate definitions of `BaseModelLimits`,
  `ModelMetadataFields`, `CachedModelMetadataSnapshot`, and
  `ResolvedModelMetadata` from `src/extension.ts` and import them from
  `src/metadata.ts` instead, so the `cost` and modality fields actually
  flow through the resolver.
- Correct the `modelCapabilities` return shape to use the official
  `vscode.LanguageModelChatCapabilities` (`imageInput`, `toolCalling`,
  `supportsImageToText`, `supportsToolCalling`) so VS Code internally
  maps them to `vision` / `toolCalling` / `agentMode` correctly.
- Use the same `In: $X · Out: $Y /1M tokens` label format that the Copilot
  extension produces in `formatPricingLabel` for visual consistency.

## Testing

- `npm run compile` clean.
- `npm test` — 12/12 pass (auth headers, stream normalizers, routing,
  usage formatting).
- Manually verified `https://models.dev/api.json` for both `opencode` and
  `opencode-go` providers — 82 models with cost data, all four buckets
  populated.

## Notes

- The cache key bump is intentional. Users on the previous build had a
  populated `v3` snapshot that didn't carry the new `cost` field; without
  the bump they would have kept seeing the placeholder labels until the
  6-hour TTL expired.
- Audio / video / PDF input support still only renders in the model
  tooltip and detail line. VS Code's "Capabilities" column is hardcoded to
  only render `Tools`, `Vision`, and `Agent Mode` based on the three
  official capability fields, so we can't promote those inputs to the
  dedicated column without an upstream VS Code change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ## Unreleased
 
+### Added
+
+- Added support for VS Code's `languageModelPricing` proposed API, exposing `pricing`, `inputCost`, `outputCost`, `cacheCost`, and `priceCategory` on every registered model so the model picker and management UI can display real cost metadata.
+- Parsed per-model cost data from the live `models.dev` registry (`cost.input`, `cost.output`, `cost.cache_read`, `cost.cache_write`) and converted USD values to AI Credits (`1 USD = 100 AI credits`) for native VS Code consumption.
+- Added modality detection from `models.dev` metadata, surfacing audio, video, and PDF input support in model tooltips and detail badges alongside the existing vision indicator.
+
+### Changed
+
+- Removed the `opencodego.experimentalContextIndicator` configuration setting and its associated context-window hook bridge; the same capability was already implemented natively in commit `ca8bbb6` and the redundant experimental path is no longer needed.
+- Consolidated duplicate local type definitions (`BaseModelLimits`, `ModelMetadataFields`, `CachedModelMetadataSnapshot`, `ResolvedModelMetadata`) that were shadowing the canonical types in `metadata.ts`, ensuring `cost` and modality fields flow correctly through the metadata pipeline.
+- Bumped the cached `models.dev` snapshot key from `v3` to `v4` so users automatically re-fetch the registry on next activation and pick up the freshly added `cost` and modality data, instead of consuming stale cached entries that did not carry those fields.
+- Aligned the `priceCategory` thresholds with the Copilot extension's 3:1 input:output weighted blend so low/medium/high/very_high buckets line up with what the user sees for the official Copilot models (e.g. Kimi k2.6 is `medium`, GPT-5.4 is `medium`, Claude Opus 4.5 is `high`, GPT-5.4 Pro is `very_high`).
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


ltmoerdani (approved):
## ✅ Approved — Great PR

Clean implementation of the `languageModelPricing` API with solid cleanup of redundant experimental code.

### 👍 What I Like
- **Pricing from `models.dev`** — real cost data surfaced natively in VS Code picker, great UX improvement
- **Type consolidation** — removing duplicate definitions from `extension.ts` and importing from `metadata.ts` fixes the shadowing bug where `cost`/modality fields weren't flowing through
- **Cache key bump `v3` → `v4`** — smart move to force re-fetch and avoid stale entries
- **`toolCalling: 128` → `true`** — fixes incorrect type, aligns with official API shape
- **Well-documented** — JSDoc on `modelPricingFields` and `costCategory` explains the "why" clearly

### ⚠️ Minor Follow-ups (non-blocking)
1. **`toolCalling: true` blanket** — not all OpenCode models support tool calling. Consider checking per-model capability to avoid showing the "Tools" badge on models that don't support it.
2. **Audio badge logic** in `formatModalityBadges` — the two conditionals for `supportsAudio` overlap and can be simplified to a single `if (metadata.supportsAudio)` check.
3. **`costCategory` thresholds** (`2`, `25`, `50`) are hardcoded — consider extracting to named constants for maintainability if pricing models change.

### 🧪 Confirmed
- CI passes (GitGuardian clean)
- `npm run compile` + `npm test` (12/12) clean per author

Merging now. Thanks @Wallacy!

PR #13, opened 2026-06-07
feat: add option to strip <think> tags from model output with configu…

### LINKED ISSUE #12: Minimax M3 <think /> blocks are displayed in the response
When using the new Minimax M3 models, the model's `<think>...</think>` content is displayed as if it were part of the response. Reasoning blocks like this should not be intermingled with the actual response, as they confuse the user and pollute the chat window.

Image

Wallacy:
### Fix `<think>` tag leakage for MiniMax M3 models

**Closes #12**

MiniMax M3 (and family) inline their reasoning content inside `<think>...</think>` tags in the regular `content` field of streaming deltas, rather than using a dedicated `reasoning_content` field. This caused the raw tag text to leak into the chat history as visible garbage.

This PR strips those tags from the output steam and routes the extracted reasoning through the existing reasoning fallback pipeline, so it displays naturally as thinking content instead of raw markup.

### Changes

| File | What |
|---|---|
| `src/streaming.ts` | Added `processThinkTagsStream()` to both `OpenAiResponseExtractor` and `AnthropicResponseExtractor` (with partial-tag buffering across chunks), `stripThinkTags()` for non-streaming fallback, `flushReasoningFallback` buffer flush, and helpers `KNOWN_INLINE_THINK_MODELS`, `shouldStripThinkTags`, `resolveStripThinkTags`. `extractChatCompletionParts` and `extractAnthropicParts` accept a `stripThinkTagsEnabled` flag. |
| `src/extension.ts` | Added `stripThinkTags` to `ApiSettings`, reads the config, passes it to all 4 streaming callers. |
| `package.json` | New `opencodego.stripThinkTags` setting (`"never"` / `"auto"` / `"always"`). |
| `CHANGELOG.md` | Entry added under `Unreleased`. |

### Design decisions

- **Only MiniMax in the known-models list** — no model is added without confirmation that it inlines `<think>` in the content field. If other models do the same, they can be added to `KNOWN_INLINE_THINK_MODELS` as they're reported.
- **Config-gated** — `opencodego.stripThinkTags` defaults to `"auto"` (apply only to known models). Set to `"always"` to strip for every model, or `"never"` to disable (e.g. if a model legitimately uses `<think>` in its output).
- **No type pollution** — everything uses `LanguageModelResponsePart[]` internally. No changes to `vscode.proposed.chatProvider.d.ts`.
- **Partial-tag safe** — `processThinkTagsStream` maintains a `thinkOpenBuffer` so `<think>` tags that split across SSE chunks are correctly handled.

### Testing

- Compilation passes (TypeScript strict, 0 errors).
- Rebased on latest `main` (v0.2.1) with clean conflict resolution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,11 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
   - Usage log persisted in VS Code `globalState` so data survives editor restarts.
   - New command: `OpenCode Go: Show Usage` (`opencodego.showUsage`).
 
+## Unreleased
+
+### Fixed
+- Strip `<think>...</think>` tags from model output when enabled. For streaming responses, the inner thinking content is accumulated into the reasoning pipeline and displayed via the existing reasoning fallback. For non-streaming fallback, the tags are removed and only the surrounding text is kept. Gated by the new `opencodego.stripThinkTags` setting (`"auto"` / `"always"` / `"never"`). In `"auto"` mode (default), stripping applies only to known models that inline reasoning in `<think>` tags within the content field (MiniMax M3 family).
+
 ## [0.1.10] — 2026-06-05
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -123,6 +123,12 @@
           "default": true,
           "description": "When enabled, only free OpenCode Zen models are shown in the Copilot model selector. Disable to include paid Zen models as well."
         },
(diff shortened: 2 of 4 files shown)


ltmoerdani:
Great PR! Clean implementation with solid edge-case handling (partial-tag buffering across SSE chunks). Tested locally and TypeScript strict compile passes with 0 errors and works as expected with MiniMax models. Let's merge this in! 🚀

PR #18, opened 2026-06-10
Fix Kimi thinking format and update documentation

Wallacy:
# Fix Kimi thinking format and update documentation

## Problem

The `opencodego.thinking.kimi` setting was sending `enable_thinking: true | false` in the request payload, but the OpenCode Go gateway **rejects** this field with HTTP 400: `"Extra inputs are not permitted"`. The correct format for MoonshotAI (Kimi) models on the chat-completions endpoint is `thinking: { type: "enabled" | "disabled" }` — the same format used by GLM.

## Root cause

The `[0.2.4]` CHANGELOG entry incorrectly documented that Kimi models should use `enable_thinking`. The extension code, however, **was never changed** to implement this — it always used `thinking: { type }`. The confusion came from a misreading of the gateway's supported formats.

### `src/extension.ts` — Thinking payload correction
- **Kimi**: `buildThinkingPayload()` now returns `thinking: { type: "enabled" | "disabled" }` instead of `enable_thinking: true | false`
- Added explanatory comment referencing tests confirming the correct format
- **GLM**: clarified comment — the gateway **does not validate/transform** GLM parameters (`variants()` returns `{}`)
- **MiniMax M3**: inline documentation clarifies that `thinking.type = "adaptive"` is the correct format (not `"enabled"`)

### `package.json` — Description fix
- `opencodego.thinking.kimi`: description updated from `enable_thinking: true | false` to `thinking: { type: 'enabled' | 'disabled' }`

## Testing

Full test battery of 70 API calls covering all 17 opencode-go models:
- **67 HTTP 200** ✅ (all thinking formats correct)
- **3 expected HTTP 400** ✅ (confirming `enable_thinking` on Kimi, `thinking.type="enabled"` on M3, and thinking on unsupported models are rejected)

No intra-family differences found.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the **OpenCode Go BYOK Provider** extension are documented here.
 
+## [Unreleased]
+
+### Fixed
+
+- **Kimi thinking format correction.** The `[0.2.4]` changelog entry incorrectly stated that Kimi models use `enable_thinking: true | false`. Tests confirm the OpenCode Go gateway **rejects** `enable_thinking` (HTTP 400: "Extra inputs are not permitted"). The correct format is `thinking: { type: "enabled" | "disabled" }` — matching GLM's format and what the gateway expects for MoonshotAI models on the OpenAI-compatible endpoint. The extension code has been using `thinking: { type }` all along; this entry corrects the record.
+
 ## [0.2.6] — 2026-06-10
 
 ### Removed
--- a/package.json
+++ b/package.json
@@ -139,7 +139,7 @@
           "type": "string",
           "enum": ["on", "off"],
           "default": "off",
-          "markdownDescription": "Thinking mode for Kimi models (`kimi-k2.5`, `kimi-k2.6`). Maps to `enable_thinking: true | false` (MoonshotAI-native boolean on the OpenAI-compatible endpoint)."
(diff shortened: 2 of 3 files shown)


Wallacy:
Sorry about the regression, when i was implementing the last PR i tested all combinations and kimi was only responding to enable_thinking: true | false ; that's why i changed... but i think that was something temporary and the old value was the correct. Next time i will do a secondary test battery with a few hours of difference when things changes like that.

ltmoerdani:
## Review: Fix Kimi thinking format

Thanks for the thorough fix and detailed PR description. LGTM overall — just one minor nit:

### ✅ What looks good
- **Bug fix is correct** — `enable_thinking` → `thinking: { type }` matches the gateway expectations and is consistent with the GLM family format
- **Testing is solid** — 70 API calls with 67 × 200 and 3 × expected 400 gives high confidence
- **Documentation updated** — `package.json` description and `CHANGELOG.md` are in sync with the code change
- **Risk is LOW** — only affects the `kimi-*` model family branch

### ⚠️ Minor nit — Language consistency
The new comment in `src/extension.ts` is in Portuguese:
```
// Testes confirmam que o gateway aceita thinking: { type } para Kimi
```
The rest of the codebase uses English comments. Would be nice to translate this to:
```
// Tests confirm the gateway accepts thinking: { type } for Kimi
```

No blocker though — feel free to address in a follow-up or leave as-is.

**Verdict: 👍 Approved to merge**

PR #38, opened 2026-06-13
fix: add option to turn off reasoning

### LINKED ISSUE #35: [BUG] Unable to Turn Off Reasoning After Selecting an Option
## 🐛 What happened?

Using Deepseek V4 Flash, when the reasoning level is selected it cannot be turned off as no corresponding option appears. In this case, the only two options are "High" and "Max". This issue however does not exist for Mimo V2.5. I have not yet tested other models.

VSCode Version: 1.124.2
Platform: macOS 26.5

sublimode:
Closes #35

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2521,10 +2521,15 @@ function buildFamilyThinkingSchema(
       const enumDescriptions: string[] = [];
 
       // "off" is always the first option when toggle is present
-      if (hasToggle) {
-        enumOptions.push("off");
-        enumLabels.push("Off");
-        enumDescriptions.push("Fastest responses");
+      enumOptions.push("off");
+      enumLabels.push("Off");
+      enumDescriptions.push("Fastest responses");
+
+      // Toggle-only (no effort values): add "on" for a simple off/on choice
+      if (hasToggle && effortValues.length === 0) {
+        enumOptions.push("on");
+        enumLabels.push("On");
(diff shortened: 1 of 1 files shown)


sublimode:
It should however be noted that I am not sure whether reasoning is being effectively controlled through these settings. VSCode chat does not seem to display reasoning tokens. May be related to https://github.com/microsoft/vscode/issues/318211

ltmoerdani:
Nice fix @sublimode! 🎉 I checked out your branch, compiled it (0 errors), traced through the logic for all scenarios, and tested manually — everything works as expected.

| Scenario | Before | After |
|---|---|---|
| Effort-only models (DeepSeek V4 Flash from models.dev: `high`, `max`) | ❌ No "off" — could not disable reasoning | ✅ `off`, `high`, `max` |
| Toggle-only models (toggle but no effort values) | ❌ Only "off" — could not enable | ✅ `off`, `on` |
| Toggle + effort models | ✅ Unchanged | ✅ Unchanged |
| Family hardcoded (DeepSeek family fallback) | ✅ Unchanged | ✅ Unchanged |

The fix is correct — moving "off" outside the `hasToggle` guard ensures it is always available, and the `hasToggle && effortValues.length === 0` condition cleanly adds "on" only when appropriate. No regressions.

This resolves the issue you reported in #35. Merging now — thanks for both reporting the bug AND fixing it! 🙏

PR #39, opened 2026-06-14
feat: show OpenCode models in Agents window (Copilot CLI) model picker

### LINKED ISSUE #11: Feature Request: Show OpenCode models in the VS Code Agents window model picker
**Requirement**

When using the **Agents window** in VS Code (starting a new Copilot CLI / Local agent session), OpenCode models should appear in the model picker alongside other providers like Claude.

Image

Image

Currently, OpenCode models registered via `vscode.lm.registerLanguageModelChatProvider()` only appear in the Chat view's model picker. They are absent from the Agents window.

**What I observed**

- The Agents window's Copilot CLI session has `requiresCustomModels: true` in its extension contribution point (package.json → `chatSessions`). This means the model picker **only** shows models that explicitly declare a matching `targetChatSessionType`.
- Models registered without `targetChatSessionType` are excluded from this picker.
- OpenRouter models appear in the Agents window, presumably because they set `targetChatSessionType: 'copilotcli'`.

**What I tested**

I attempted to add `targetChatSessionType` to the registered models in `provideLanguageModelChatInformation()`, testing the following values:

| Value | Result |
|---|---|
| `'copilotcli'` | Did not appear in the Agents window picker |
| `'agent-host-copilot'` | Caused strange UI behavior (extra options appearing) |
| `'agent-host-copilotcli'` | The Copilot CLI session uses resource scheme `agent-host-copilotcli` — models with this value appeared in the picker |

**Related VS Code PR**

[PR #298862 — "Check language model session target"](https://github.com/microsoft/vscode/pull/298862) clarifies that models with `targetChatSessionType` are scoped exclusively to their targeted session type and excluded from `getDefaultLanguageModel()`.

**Relevant VS Code source constants**

From `src/vs/workbench/contrib/chat/common/chatSessionsService.ts`:
- `SessionType.CopilotCLI = 'copilotcli'`
- `SessionType.AgentHostCopilot = 'agent-host-copilotcli'`

From `src/vs/workbench/contrib/chat/browser/agentSessions/agentSessions.ts`:
- `AgentSessionProviders.Background = SessionType.CopilotCLI`

Marinski:
## Summary

Fixes #11 - OpenCode models now appear in the VS Code Agents window model picker when using the Copilot CLI agent session.

## Problem

Models registered via `vscode.lm.registerLanguageModelChatProvider()` were visible in the Chat view's model picker but completely absent from the Agents window (Copilot CLI session). The root cause: VS Code's model picker for the Copilot CLI session only shows models that explicitly declare a matching `targetChatSessionType`, and our models had none.

## Solution

Register each model **twice** from `provideLanguageModelChatInformation()`, using `flatMap` instead of `map`:

| Variant | `id` | `targetChatSessionType` | Visible in |
|---|---|---|---|
| General | `opencodego:deepseek-v4-flash` | _(none)_ | Chat view |
| Agents-window | `opencodego:deepseek-v4-flash::agent-host` | `"copilotcli"` | Agents window ? Copilot CLI session |

### No picker duplication

VS Code's `filterModelsForSession()` logic partitions the two variants - each surface sees only the entries meant for it. This is confirmed by [microsoft/vscode#298862](https://github.com/microsoft/vscode/pull/298862), which makes `getDefaultLanguageModel()` also exclude session-targeted models. Neither picker shows both variants.

### Routing is unchanged

The `::agent-host` suffix in the agent-window model ID is stripped by the existing `resolveRawModelId()` helper (splits on `::`, takes the first segment), so all backend routing, API key lookup, and metadata resolution continue to use the real model ID.

### `targetChatSessionType` value

The correct value is `"copilotcli"` - the `type` field from the Copilot extension's `chatSessions` contribution in its `package.json`:
`json
{ "type": "copilotcli", "requiresCustomModels": true, ... }
`
Earlier research pointed to `"agent-host-copilotcli"` (the VS Code internal resource URI scheme), which is different and does not work.

## Required user setting

The Agents window does not activate third-party extensions by default. Users need to add this to their VS Code `settings.json`:

`json
"extensions.supportAgentsWindow": {
    "ltmoerdani.opencode-copilot-chat": true
}
`

## Testing

1. Add the `extensions.supportAgentsWindow` setting above to VS Code user settings.
2. Reload the Agents window.
3. Start a new session ? select **Copilot CLI** as the agent type.
4. Open the model picker - OpenCode Go and OpenCode Zen models appear alongside Claude / GPT models.
5. Verify the Chat view model picker shows each model exactly once (no `::agent-host` entries visible there).

## Files changed

- `src/extension.ts` - `provideLanguageModelChatInformation()` changed from `map` to `flatMap`, returning `[generalVariant, agentsWindowVariant]` per model.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1128,13 +1128,15 @@ class OpenCodeProvider implements vscode.LanguageModelChatProvider<OpenCodeModel
     const settings = getSettings();
     const metadataSnapshot = await this.getMetadataSnapshot();
 
-    return models.map((modelId) => {
+    return models.flatMap((modelId) => {
       const metadata = this.resolveModelMetadata(modelId, metadataSnapshot);
       const routing = resolveModelRouting(modelId, this.definition);
       const effectiveModelId = toEffectiveModelId(modelId, this.definition.vendor);
+      const agentHostModelId = `${effectiveModelId}::agent-host`;
       const limits = modelLimits(metadata, settings);
       this.apiKeysByModelId.set(modelId, apiKey);
       this.apiKeysByModelId.set(effectiveModelId, apiKey);
+      this.apiKeysByModelId.set(agentHostModelId, apiKey);
 
       const capacityNote = CAPACITY_LIMITED_MODEL_NOTES[modelId];
       const modalityBadges = formatModalityBadges(metadata);
@@ -1142,8 +1144,7 @@ class OpenCodeProvider implements vscode.LanguageModelChatProvider<OpenCodeModel
       const baseTooltip = `${this.definition.displayName} model: ${modelId}`;
       const configurationSchema = modelConfigurationSchema(modelId, metadata);
 
-      const info: OpenCodeModel = {
-        id: effectiveModelId,
+      const sharedFields: Omit<OpenCodeModel, "id" | "targetChatSessionType"> = {
         rawModelId: modelId,
         name: `${this.definition.modelNamePrefix} / ${formatModelName(modelId)}`,
         family: `${this.definition.vendor}-${modelId}-${MODEL_METADATA_REVISION}`,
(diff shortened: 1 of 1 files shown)


ltmoerdani:
Thanks @Marinski! Great implementation, tested locally, built the VSIX, installed, and verified working. ✅

**Verification results:**
- ✅ `tsc` compile: 0 errors
- ✅ Build + install VSIX: models appear in Agents window (Copilot CLI session) model picker
- ✅ Chat view model picker: each model still shows exactly once (no `::agent-host` duplication)
- ✅ Routing verified: `::agent-host` suffix correctly stripped by `resolveRawModelId()`, API key lookup and model resolution all work
- ✅ Marketplace-safe: no `enabledApiProposals` needed

Matches the Option A solution documented in `docs/references/01-20260611-agents-window-model-visibility.md`.

**One minor note (non-blocking):** `showDiagnostics()` will now list each model twice because `selectChatModels()` returns both variants. Purely cosmetic in the diagnostics output, can be addressed separately if needed.

Merging now. 🚀

PR #42, opened 2026-06-15
fix: gate agent-host model duplicate behind opt-in setting (fixes #41)

### LINKED ISSUE #41: [BUG] Models are shown twice (once with agent-host)
## 🐛 What happened?

All models are shown twice.

Image

---

Model id/name (any model suffers from this, not just qwen):

Those are the 2 models displayed:
`opencodego:qwen3.7-max::session-2026-05-21-b`
`opencodego:qwen3.7-max::session-2026-05-21-b::agent-host`

## ✅ What did you expect?

See each model only once.

## 🖥️ Environment

- **Extension version:** v0.3.0
- **VS Code version:** v1_124

Marinski:
## Problem

PR #39 introduced a regression: every model was registered twice - a general variant and a `::agent-host` variant with `targetChatSessionType: "copilotcli"`. While `filterModelsForSession()` correctly hides the `::agent-host` copy from the Chat view dropdown and the Agents window, the **Language Models management UI** (the BYOK enable/disable list) enumerates the raw registration list with no session filtering - so both variants appeared there, showing every model twice with a `::agent-host` suffix (reported in #41).

## Fix

Gate the `::agent-host` duplicate behind a new **opt-in** setting `opencodego.showInAgentsWindow` (default `false`).

### Behaviour

| Setting | Chat view | Agents window | Management UI |
|---|---|---|---|
| `false` (default) | ? one entry per model | ? no OpenCode models | ? one entry per model |
| `true` | ? one entry per model | ? OpenCode models visible | Two entries: `Model Name` + `Model Name (Agents)` |

When `showInAgentsWindow: true`, the `::agent-host` variant gets an `(Agents)` name suffix so the two entries are visually distinguishable in the management UI.

### Required additional setting (unchanged from PR #39)

`json
"extensions.supportAgentsWindow": {
    "ltmoerdani.opencode-copilot-chat": true
}
`

## Files changed

- `src/extension.ts` - reads `opencodego.showInAgentsWindow`; returns `[info]` by default, `[info, agentHostInfo]` when opted in; adds `(Agents)` name suffix to the agent-host variant.
- `package.json` - adds `opencodego.showInAgentsWindow` boolean setting (default `false`) with full description.

Fixes #41

### DIFF
--- a/package.json
+++ b/package.json
@@ -140,6 +140,11 @@
           "default": true,
           "description": "When enabled, only free OpenCode Zen models are shown in the Copilot model selector. Disable to include paid Zen models as well."
         },
+        "opencodego.showInAgentsWindow": {
+          "type": "boolean",
+          "default": false,
+          "markdownDescription": "Also expose OpenCode models in the Agents window (Copilot CLI session) model picker by registering each model a second time with `targetChatSessionType: \"copilotcli\"`. **Disabled by default** to keep the model list clean — when disabled each model appears exactly once everywhere. When enabled, an `(Agents)` suffix is added to the duplicate entry so the two are distinguishable in the Language Models management UI. Also requires `extensions.supportAgentsWindow: { \"ltmoerdani.opencode-copilot-chat\": true }` to activate this extension in the Agents window."
+        },
         "opencodego.stripThinkTags": {
           "type": "string",
           "enum": ["never", "auto", "always"],
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1128,6 +1128,16 @@ class OpenCodeProvider implements vscode.LanguageModelChatProvider<OpenCodeModel
     const settings = getSettings();
     const metadataSnapshot = await this.getMetadataSnapshot();
 
+    // The Agents-window variant (targetChatSessionType: "copilotcli") is only
+    // hidden from the Chat view's in-session dropdown by filterModelsForSession().
+    // Surfaces that enumerate the raw registration list — most notably the Language
+    // Models management UI — show ALL registered models with no session filtering,
+    // so the ::agent-host copy appeared there too (issue #41, regression from PR #39).
+    // Gate the duplicate behind an opt-in setting so the default stays clean.
+    const showInAgentsWindow = vscode.workspace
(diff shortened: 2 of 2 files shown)


ltmoerdani:
Thanks @Marinski — solid fix. Verified the diff against local:

✅ **Correct root cause.** `filterModelsForSession()` hides the `::agent-host` variant from the Chat view dropdown and Agents window picker, but the Language Models management UI (BYOK enable/disable list) enumerates the raw registration list with no session filter — so both variants leaked there (#41, regression from #39). Gating behind an opt-in default-off setting is the right call: restores the pre-#39 clean list by default while keeping the Agents-window feature available for users who want it.

✅ **Implementation looks clean:**
- Setting read via standard `getConfiguration("opencodego").get("showInAgentsWindow", false)` — matches the pattern used by `freeOnly` / `showUsageStatusBar`.
- Default `false` is the correct regression fix (clean list everywhere by default).
- `(Agents)` name suffix on the duplicate entry keeps the management UI readable when opted in, without touching routing — `::agent-host` suffix in `id` is still stripped by `resolveRawModelId()`, so backend routing / API key lookup / metadata resolution are unchanged.
- `apiKeysByModelId.set(agentHostModelId, …)` still runs even when gated off — safe (no dangling key lookup if the user toggles the setting on later without a reload).
- Log statement moved before the early return, so the output channel stays clean when opt-in is off.
- No `enabledApiProposals` touched → marketplace-safe.

Minor heads-up for users upgrading from v0.3.0: if they were relying on OpenCode models in the Agents window, they'll need to **also** set `"opencodego.showInAgentsWindow": true` now (in addition to `extensions.supportAgentsWindow`). I'll cover this in the CHANGELOG + README update when cutting the release, and update `docs/features/06-20260614-agents-window-model-visibility.md` with a "Post-#42: opt-in gating" section so the history stays consistent.

Nice touch preserving @Wallacy's alternative (separate providers) as context in #41 — went with the minimal-risk setting approach here, which I agree with for a hotfix.

CI is green (2/2). Will merge with a merge commit to preserve your commit. 🚀

PR #50, opened 2026-06-16
feat: manual usage targets + live pricing for Go tracker

### LINKED ISSUE #23: OpenCode Go - Usage: status not updating
How often is the usage stats card supposed to be updating because it doesn't seem to be.  Most obvious because the monthly reset shows 27d left where it should be about 19d according to opencode.ai's stats (I happened to start my subscription on 6/1).  

Compared to the opencode website, the following are the 3 discrepancies:
Weekly Usage % is incorrect (9% too low)
Monthly Usage % and time are incorrect (15% too low, and should be 19d left)

Image

I have updated VS Code & rebooted a couple times in the last week.

Wallacy:
## Summary

Adds manual Go usage target configuration and fixes cost estimation to use live models.dev pricing instead of a stale hardcoded table.

## Closes

Closes #23

### Manual Usage Targets (`showUsageTargetEditor`)
- New command **"OpenCode Go: Set Usage Targets..."** opens 5 sequential input boxes pre-filled with current values:
  - Session spent (rolling 5h, limit $12)
  - Weekly spent (Mon–Mon UTC, limit $30)
  - Monthly spent (limit $60)
  - Monthly reset **day** (1–31)
  - Monthly reset **hour** (0–23 UTC)
- Press Enter on any input to keep the current value; Escape cancels the entire flow.
- Targets are persisted in `globalState` as baseline amounts with a custom `expiresAt` for the monthly anchor.

### Monthly Reset Display Fix
- `buildSummaryFromTracked` and `buildSummaryFromRows` now use `baseline.monthly.expiresAt` as `resetsAt` when available, so the "resets in Xd Yh" text reflects the user-configured anchor date instead of the auto-calculated month end.

### Webview / Status Bar Cleanup
- Removed button and scripts from the usage webview (now display-only SVG).
- Status bar click disabled (hover-only via tooltip).
- Tooltip links to the `setUsageTargets` command.

### Live Pricing Fallback
- Injected a `CostResolver` callback into `GoUsageTracker` that reads from the live `models.dev` metadata snapshot.
- Cost resolution priority: per-request `externalCost` → live snapshot → bundled static table.
- The hardcoded `GO_MODEL_PRICING` is kept only as a last-resort fallback.

## Commits

- `d031457` — feat: add manual usage targets editor for Go usage tracker
- `094978e` — fix: use live models.dev pricing instead of hardcoded fallback in Go tracker

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,10 @@
       {
         "command": "opencodego.showUsageDetails",
         "title": "OpenCode Go: Show Usage Details"
+      },
+      {
+        "command": "opencodego.setUsageTargets",
+        "title": "OpenCode Go: Set Usage Targets…"
       }
     ],
     "configuration": {
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -55,6 +55,7 @@ import {
 import {
   GoUsageTracker,
   formatGoUsageStatusBarText,
+  type UsageBaselineTargets,
 } from "./goUsageTracker";
 
 const SECRET_KEY = "opencodego.apiKey";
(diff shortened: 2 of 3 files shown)


ltmoerdani:
@Wallacy, good work. Implementation matches the description (I checked against the diff), and the monthly reset fix lands in both `buildSummaryFromRows` and `buildSummaryFromTracked`, so neither code path is left inconsistent. Keeping `GO_MODEL_PRICING` as a fallback is the right call: the extension has to stay functional when live fetch fails. `enableScripts: false` is a solid security tightening. Monthly anchor rollover handles the "next month if already passed" edge case correctly.

CI is green. Three things worth noting before or after merge, your call:

1. `setCostResolver()` is dead code. The `CostResolver` is already captured via constructor closure from `modelMetadataSnapshot`, so the setter is never called. Either wire it up after metadata refresh, or remove it to keep the API surface honest.

2. Validation limits in `showUsageTargetEditor` (`$12`, `$30`, `$60`) duplicate `GO_LIMITS`. If those limits change, validation silently goes stale. Importing `GO_LIMITS` directly fixes the drift.

3. `md.isTrusted = true` without `supportedCommands`: the tooltip link `[$(pencil) Set spent targets](command:opencodego.setUsageTargets)` may render as plain text in some VS Code versions. Adding `md.supportedCommands = ["opencodego.setUsageTargets"]` makes the intent explicit. Worth a quick hover test.

Merging now.

PR #52, opened 2026-06-17
fix: baseline negative support, status bar QuickPick, input validation, tooltip refresh

Wallacy:
## Summary

Fixes baseline editing bug where session/weekly targets silently failed on re-edit, adds status bar click → QuickPick for usage details, and improves input validation.

### Fixed
- **Baseline editing now correctly recalculates on re-edit.** `setManualSpentTargets()` now computes `baseline = target - tracked` (without clamping to 0) for all three periods. This allows **negative baselines** that offset tracked entries downward, so `display = tracked + baseline = target` exactly. Previously, `Math.max(0, ...)` prevented negative baselines — when a user lowered a target below the tracked amount, the display never updated.
- **Input validation rejects non-numeric characters.** The usage target editor now validates input with a strict regex (`/^-?\\d+[.,]?\\d*$/`) that rejects strings like `60f` or `18asd`. Accepts both `.` and `,` as decimal separators for international users.
- **Status bar tooltip refreshes immediately after editing targets.** `refreshGoUsageStatusBar()` is now called right after `setManualSpentTargets()` so the hover tooltip updates without requiring a VS Code window reload.

### Added
- **Status bar click opens usage QuickPick.** Clicking the OpenCode Go usage status bar item now opens a QuickPick with session/weekly/monthly progress bars, usage percentages, and reset countdowns. Includes quick actions to \"Set spent targets…\" and \"Open full usage panel\".

## Testing

1. Set session/weekly/monthly targets → verify display shows exact values
2. Re-edit targets → verify values change (not sum)
3. Click status bar → QuickPick opens with usage summary
4. Verify tooltip updates immediately after editing targets
5. Try entering invalid input (e.g. `60f`, `18asd`) → should be rejected
6. Try entering `3,50` → should be accepted as `3.50`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,12 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ## [Unreleased]
 
-_No changes yet._
+### Fixed
+
+- **`[Usage]` Session/weekly targets now apply correctly.** Previously, editing session or weekly spent values could silently fail when the baseline had already been zeroed by a prior edit (the delta calculation resulted in a negative value clamped to 0). Session and weekly baselines are now set to the absolute target value, so editing always reflects the user's input. Monthly continues to use delta-based calculation to preserve the anchor expiry logic.
+- **`[Usage]` Removed dead `setCostResolver()` method.** The `CostResolver` is injected via constructor closure and never updated after initialization — the setter was unreachable code.
+- **`[Usage]` Validation limits now derive from `GO_LIMITS`.** Input box validation for session ($12), weekly ($30), and monthly ($60) limits now reads from the exported `GO_LIMITS` constant instead of hardcoded numbers, preventing drift if limits change.
+- **`[Usage]` Tooltip command link now uses `supportedCommands`.** Added `md.supportedCommands = ["opencodego.setUsageTargets"]` to the usage tooltip so the `[$(pencil) Set spent targets]` link renders as a clickable command in all VS Code versions.
 
 ## [0.3.3] — 2026-06-17
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,15 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


ltmoerdani (approved):
Thanks, Wallacy. PR approved, with one minor non-blocking note at the end.

### ✅ What's Correct

**1. Negative baseline fix (`goUsageTracker.ts`):** Root cause is right. `Math.max(0, ...)` was clamping baseline at 0 whenever the user lowered a target below the tracked amount, so the display froze. Allowing negative baselines restores the invariant: `display = tracked + baseline = target`. The comment about `expiresAt` expiry is accurate too.

**2. Status bar QuickPick:** Clean. `buildUsageQuickPickItems()` is reused from `goUsageTracker.ts:679`, and both dispatched commands (`setUsageTargets`, `showUsageDetails`) were already registered. Nothing invented.

**3. `GO_LIMITS` as single source of truth:** Good call. Prevents validation logic and the actual limits from drifting apart over time.

**4. `refreshGoUsageStatusBar()` after setting targets:** Small fix, real impact. Previously required a full window reload.

**5. Dead code removal of `setCostResolver()`:** Correct. `CostResolver` comes in via the constructor closure and was never updated externally. Verified no callers.

### ⚠️ Minor Notes (non-blocker)

**The `(md as any)` cast in `buildUsageTooltip`:**

```ts
(md as any).supportedCommands = ["opencodego.setUsageTargets"];
```

The codebase enforces a strict no-`any` rule. This is a proposed API not present in VS Code's stable types, which is why the cast was used — but a cleaner form would be:

```ts
(md as unknown as { supportedCommands?: string[] }).supportedCommands = ["opencodego.setUsageTargets"];
```

Or add it to `vscode.proposed.chatProvider.d.ts`. Not a merge blocker; fine as a follow-up.

**`parseCurrencyInput` regex is slightly loose:** `/^-?\d+[.,]?\d*$/` accepts `"3."` (trailing separator with no digits). Not a bug since `parseFloat("3.")` still returns 3, but could be tightened to `^-?\d+(?:[.,]\d+)?$` if you want stricter validation. Optional.

🚢 **LGTM** — ready to merge with a merge commit.

PR #53, opened 2026-06-22
fix: resolve model picker crash (#51) on VSCode 1.126

### LINKED ISSUE #51: [BUG] Spamming with notifications: Could not fetch OpenCode Go (Agents) model list
## 🐛 What happened?

```
Could not fetch OpenCode Go (Agents) model list. Using bundled model list. fetch failed
```

Wallacy:
### Problem

The chat model picker crashes on VS Code 1.126+ before it opens, and OpenCode models don't appear in the picker or Agents window.

### Changes

1. **`category` type fix** — Changed from `{ label, order }` (object) to plain `string`. The unified picker calls `category.charAt(0)` which threw `TypeError` on an object. Fixes #51.

2. **Secrets fallback via `options.configuration`** — VS Code 1.126 sends `options.configuration={}` (empty object) instead of `{ apiKey: "..." }` for non-agent providers. The extension now falls back to `SecretStorage` when configuration is present but contains no usable API key:

```typescript
if (!apiKey && (this.definition.isAgentVariant || options.configuration)) {
  apiKey = await context.secrets.get(SECRET_KEY);
}
```

   - `configuration=undefined` → VS Code is still resolving → return `[]`
   - `configuration={apiKey:"..."}` → BYOK key resolved → use it
   - `configuration={}` → VS Code 1.126+ empty config → secrets fallback
   - Agent variants → always secrets (never receive BYOK keys)

   Works across all VS Code versions without any version checks.

3. **Debug log** — Added `[picker]` log showing `options` content for debugging picker behavior across versions.

4. **README / CHANGELOG updates** — Documented Agent Window behavior and model provider tables.

### Testing

- [x] VS Code 1.125: models appear in chat picker, no duplication
- [x] VS Code 1.126: models appear in chat picker, no duplication
- [x] Agent variants appear in Agents window on both versions
- [x] Compiles cleanly, 40/40 unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,12 +10,19 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ### Fixed
 
+- **`[Model Picker]` Chat model picker no longer crashes on VS Code ≥1.126.** The `category` field in `provideLanguageModelChatInformation` was typed as `{ label, order }` (object), but VS Code's `LanguageModelChatInformation.category` expects a plain `string`. The unified picker calls `category.charAt(0)` → `TypeError` on an object. Changed to a plain string (`this.definition.displayName`). Fixes [#51](https://github.com/ltmoerdani/opencode-copilot-chat/issues/51).
+- **`[Model Picker]` OpenCode models now appear on VS Code ≥1.126.** VS Code 1.126 no longer passes `options.configuration` (BYOK key) to non-agent providers, so the extension falls back to `SecretStorage` when the BYOK config is not available. This is version-gated via `vscode.version` to avoid breaking the two-call resolution on 1.125.
 - **`[Usage]` Baseline editing now correctly recalculates on re-edit.** The `setManualSpentTargets()` function now computes `baseline = target - tracked` (without clamping to 0) for all three periods. This allows **negative baselines** that offset tracked entries downward, so `display = tracked + baseline = target` exactly. Previously, `Math.max(0, ...)` prevented negative baselines — when a user lowered a target below the tracked amount, the display never updated because the delta clamped to 0. Also, session and weekly now use delta-based calculation (same as monthly) instead of absolute target injection, ensuring consistency and proper expiry behavior.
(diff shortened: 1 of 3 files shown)


ltmoerdani:
Thanks @Wallacy. Root cause analysis is spot on, and the fixes look correct. Checked against the bundled `vscode.proposed.chatProvider.d.ts` — `category` isn't declared as `{ label, order }` there, only `priceCategory?: string`, so the object was never type-correct to begin with.

Two quick questions before I merge:

1. The API key fallback for non-agent providers now also reads from `secrets`. Before, only agent variants fell back to secrets; non-agent providers only honored the BYOK config key. Side effect: clearing the key via Manage Models won't hide the models if a stale key lingers in `secrets`. Intentional? If yes, a one-line comment would help.

2. `agentProvider.triggerChange()` is now called unconditionally (previously only when the key changed). Can you confirm it's debounced or idempotent? Otherwise the provider path could cycle: `triggerChange` → `provideLanguageModelChatInformation` → `triggerChange`.

Minor (not blocking): `categoryOrder` in `ProviderDefinition` is now dead code. Happy to strip it in a follow-up PR.

Wallacy:
Im got some models duplicated here... lets me fix this first before merge.

edit: Branch still not good... 1.126 still duplicate the entries, because the secret store failback... but whiout it dosent show on the model picker.

Wallacy:
Okay @ltmoerdani i think that i got this. Do a double check plz.

ltmoerdani:
@Wallacy The new approach reads better. The `options.configuration` truthy check is the right discriminator here: empty `{}` from 1.126 triggers the secrets fallback, `undefined` lets VS Code resolve again without duplicating. No version sniffing, no group guard needed.

That answers both of my earlier questions. The stale-key edge case in secrets is still theoretically there, but the discriminator narrows the blast radius enough that I'm comfortable shipping it.

Compiled clean on my end (`tsc -p ./`, zero errors). Merging now with `--merge` to keep all commits intact.

PR #54, opened 2026-06-23
chore: security hardening and optimization cleanup

Wallacy:
## chore: security hardening and optimization cleanup

Follow-up to #53 — addresses security concerns, dead code, and a memory
leak identified during a full review of the extension's security posture.

### Security

**1. Picker debug log leaked API keys to the Output channel (critical)**

The previous `provideLanguageModelChatInformation` logged `JSON.stringify(options)` on every picker resolution. When VS Code sent `options.configuration={apiKey:"sk-..."}`, the full API key was written to the OpenCode Output channel in plaintext — visible to anyone with access to the user's machine. This was a temporary diagnostic added during the 1.125/1.126 picker investigation and should never have shipped. **Removed entirely.**

**2. `Clear API Key` didn't fully clear the key (medium)**

When a user cleared the key via the extension's "Manage Models" → "Clear API Key" action, only `SecretStorage` was wiped. The next picker resolution re-stored the key from `options.configuration` (BYOK) into `SecretStorage`, making the clear appear to have no effect. The info message now warns the user to also remove the key from VS Code's Manage Models panel for a full clear. (This is a UX improvement — the underlying behavior is correct by design since BYOK is the primary key source.)

### Memory Leak

**3. `reasoningContentByToolCallId` grew without limit (long sessions)**

Each tool call in an agent session added one entry to this `Map` (toolCallId → reasoning text). Entries were never evicted. Over a multi-hour agent session with hundreds of tool calls, this could accumulate thousands of entries of reasoning content (typically 200–500 tokens each). 

The fix caps the map at 500 entries. When exceeded, the oldest entries (by insertion order) are evicted. At ~200–500 tokens per reasoning chunk, 500 entries ≈ 100K–250K tokens of cached reasoning — comfortably covering even the most intensive agent workflows (e.g., a session that processes 500+ tool calls would only lose the earliest reasoning chunks, which are no longer needed by the time the map fills). The cap is implemented in a new `storeReasoningContent()` helper method.

### Dead Code Removal

**4. `agentProvidersByBaseVendor` map — removed**

This `Map<string, OpenCodeProvider>` was populated during activation to hold references to agent-variant provider instances. It was originally used by `triggerChange()` to force agent providers to re-resolve after the base provider stored the BYOK key. After #53, agent variants resolve independently via the secrets fallback, making `triggerChange()` unnecessary (it was already removed). The map was left behind — it held strong references to provider instances for no reason.

**5. `categoryOrder` field — removed from `ProviderDefinition`**

This field was a remnant of the old `category: { label, order }` object that crashed the picker on VS Code ≥1.126 (fixed in #53). After the category was changed to a plain string, `categoryOrder` was never read. It also polluted `providerVariant()` with an unused parameter.

### Testing

- [x] Extension compiles cleanly
- [x] 40/40 unit tests pass
- [x] No behavioral changes — all fixes are internal (security, memory, dead code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ## [Unreleased]
 
+### Changed
+
+- **`[Model Picker]` Removed `triggerChange()` cross-provider re-resolution.** The agent-variant providers now resolve their API key independently via the secrets fallback path (`isAgentVariant || options.configuration`). Previously, the base provider called `triggerChange()` on the agent provider to force it to re-resolve after storing the BYOK key — this indirection is no longer needed since both paths reach the same `SecretStorage` read.
+- **`[Model Picker]` Removed `categoryOrder` from `ProviderDefinition`.** This field was left over from the old `category: { label, order }` object that crashed the picker on VS Code ≥1.126 (fixed in 0.3.4). It was never read after the type was changed to a plain string.
+
+### Fixed
+
+- **`[Security]` Picker debug log no longer leaks API keys.** The previous `JSON.stringify(options)` debug log wrote the full `options` object (including `configuration.apiKey`) to the Output channel in plaintext. Removed entirely — the log served only as a temporary diagnostic during the 1.125/1.126 picker investigation and is no longer needed.
+- **`[Security]` `Clear API Key` action now warns about BYOK re-persistence.** When the user clears the key via `SecretStorage`, the next picker resolution re-stores it from `options.configuration` (BYOK). The info message now tells the user to also remove it from the Manage Models panel if they want a full clear.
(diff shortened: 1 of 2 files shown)


ltmoerdani:
@Wallacy Picked through each claim against `main` this morning. Five for five, all real.

The API key log leak is the big one. `JSON.stringify(options)` was writing `configuration.apiKey` to the Output channel on every picker resolution. You're right that it was a leftover from the 1.125/1.126 investigation, but it should not have shipped in 0.3.4. Glad you caught it.

Quick notes on the rest:

1. `reasoningContentByToolCallId` was genuinely unbounded. Three `set()` sites, no eviction. The new `storeReasoningContent()` cap at 500 is reasonable; at roughly 200 to 500 tokens per reasoning chunk, that's somewhere between 100K and 250K tokens cached before the oldest entries roll off. More headroom than any agent session I've seen.
2. `agentProvidersByBaseVendor` only existed to feed `triggerChange()`. With agent variants now resolving through the secrets fallback (#53), the map held strong references to nothing. Clean cut.
3. `categoryOrder` is the follow-up I flagged on #53. All five references were write-only, zero reads. Thanks for finishing it.

One UX note on the Clear API Key warning: the underlying behavior is correct (BYOK is the primary source, secrets is the cache), so the message is the right fix rather than changing the logic. Worth keeping in mind if anyone files a "clear doesn't work" issue later.

Compiled clean on my end (`npm run compile`, zero errors). I have not manually tested the picker on 1.125/1.126 yet, so I'll hold the merge until I can, or you can confirm agent variants still resolve after dropping `triggerChange()`.

Will merge with `--merge` to keep both commits intact.

PR #55, opened 2026-06-24
feat: session-level cost tracking — feature parity with Copilot Chat

Wallacy:
### Goal

Bring full cost transparency and feature parity with Copilot Chat to OpenCode Go usage, matching what Copilot Chat users see: per-session cost in the status bar, tooltip, icon menu, and (eventually) the native VS Code session info popover.

### What was implemented

**Session cost tracking** — each chat thread now accumulates its cost, request count, and token usage. Visible in:

- **SVG hover card** (status bar tooltip): `Session: $0.0042 · Requests: 3 · Tokens: 4.2K`
- **QuickPick (icon menu)**: `💬 This Session` item in the Daily Summary section
- **Usage webview**: same session info in the SVG card
- Persisted across restarts via `globalState` (max 50 sessions, idle >2h pruned)

**VS Code `copilotCredits` integration** — added `copilotCredits` to the usage payload (`UsageSnapshot`, `ProviderUsagePayload`, `TransportRequestSummary`, `UsageLogEntry`). The `onTransportSummary` callback computes credits as `cost × 100` (1 credit = $0.01) using the same pricing fields already returned in `LanguageModelChatInformation`. This follows the exact same pattern as Copilot's own BYOK providers (`AnthropicLMProvider`, `GeminiNativeProvider`), which also emit a `LanguageModelDataPart({ copilot_credits }, "usage")` at the end of each response.

### Why it doesn't appear in VS Code's native session info popover (Known Issue)

VS Code 1.126 accumulates session cost by reading `usage.copilotCredits` from `IChatUsage` progress events (`{ kind: 'usage', copilotCredits: ... }`). For the `copilot` vendor this works because the `ToolCallingLoop` explicitly calls `stream.usage({ copilotCredits })` after each model fetch.

For all BYOK providers (including OpenCode), the extension already emits `LanguageModelDataPart({ copilotCredits: 0.42 }, "usage")` — the standard mechanism. However, VS Code 1.126's `ChatService.acceptResponseProgress()` **does not convert** `{ type: 'data', mimeType: 'usage' }` from BYOK provider streams into `IChatUsage` events. This is a VS Code-side limitation, not a missing extension feature. The data is correctly structured and available; the plumbing simply stops at the `ChatService` boundary for non-Copilot vendors.

We expect this to be addressed in a future VS Code release. Until then, session cost is fully visible in the extension's own status bar tooltip and icon menu.

### Changelog

```
Added:
- Session-level cost tracking (tooltip, QuickPick, SVG, persistence)
- VS Code session cost fields (copilotCredits) in usage payload

Known Issues:
- Session cost not shown in VS Code native session info (BYOK limitation)
```

### Testing
- [x] Compiles, 40/40 unit tests pass
- [x] Session cost in tooltip, QuickPick, webview
- [x] Data survives VS Code restart
- [x] No regression on existing subscription-level tracking


ltmoerdani:
@Wallacy nice one. The `copilotCredits` plumbing matches what Copilot's own BYOK providers do, and moving `onTransportSummary` above `emitSummary` is the right call. I pulled the branch and ran `npm run compile`: clean. The Known Issue note on `acceptResponseProgress` is accurate too.

A few things before I merge. I checked each against the actual branch code.

**1. Cost gets computed twice**

In `extension.ts`, `onTransportSummary` computes cost inline and mutates `summary.copilotCredits`. Then `goUsageTracker.record()` runs `estimateCost()` again and writes its own `copilotCredits` onto the `UsageLogEntry`. So the value shipped to VS Code and the value persisted in the tracker can drift the moment either formula changes.

`estimateCost()` already handles the `cache_read ?? input * 0.1` fallback. Pull that into a shared helper (`computeCredits(cost, prompt, completion, cached)`?) and call it once?

**2. No tests for the new session logic**

"40/40 unit tests pass" is the existing suite. Nothing covers the new code:

- `getCurrentSessionCost()` with multiple sessions
- `pruneSessions()` hitting the 2h idle cutoff and the 50-session cap
- `record()` accumulating `cost +=` and `requests++`
- Restore from `globalState`

`src/test/` already has the pattern (`metadata.test.ts`, `retry.test.ts`, `thinking.test.ts`). A `goUsageTracker.test.ts` for accumulation and pruning would cover the bits most likely to silently break.

**3. The session cost shown is an estimate, not the billed number**

`getSummary()` returns `buildSummaryFromTracked(...)`, the local estimate. `buildSummaryFromRows` (the SQLite reader) is dead code right now: only the unused `hasSQLiteData` getter calls it. So the tooltip shows the local number, which can drift from what OpenCode actually bills because server-side cache affinity isn't visible to the extension (see issue #23).

Either a small "(estimated)" tag in the tooltip, or a follow-up to wire up the SQLite reader. Your call.

**4. "💬 This Session" is misleading with multiple threads**

`getCurrentSessionCost()` returns the globally most-recent session, not the thread the user is looking at. Two panels open, click the icon from Thread A, you still see Thread B's cost. Rename to "Latest Session", or thread the active `sessionId` through when we have it.

**5. `usage.ts` is missing a trailing newline**

```diff
+      : { copilotCredits: normalized.copilotCredits }),
   };
}\ No newline at end of file
```

Two-second fix, saves diff noise next time someone touches the file.

The VS Code limitation writeup is the best part of this PR. Address #1 and #2 and I'll merge. The rest can ride as follow-ups.

Wallacy:
All 5 review points addressed in the latest push:

**1. Cost duplication** — Extracted `estimateCost()` as a shared export from `goUsageTracker.ts`. The `onTransportSummary` callback now calls it instead of inlining the formula, so both the usage data part (`copilotCredits`) and the tracker use the exact same code path.

**2. Tests** — Added 35 unit tests in `src/test/goUsageTracker.test.ts` covering `estimateCost()`, `record()` accumulation, `getCurrentSessionCost()` / `getRecentSessionCosts()`, state restoration from `globalState`, idle session pruning, the 50-session cap, and edge cases (zero tokens, unknown models, missing sessionId).

**3. Estimated value marker** — Since the SQLite history only contains `createdMs` and `cost` (no modelId, token counts, or sessionId), the local estimate is the only viable approach for per-session tracking. Added `(est)` suffix to both the SVG tooltip (`Session (est):`) and QuickPick label (`Latest Session (est)`) so users know it's an approximation. Also widened the SVG card (330 → 345 px) when session data is present to fit the longer label.

**4. Label ambiguity** — Renamed from `💬 This Session` to `💬 Latest Session (est)` to avoid implying it's scoped to the viewer's active thread.

**5. Trailing newline** — Fixed `usage.ts`.

ltmoerdani:
All five landed, thanks @Wallacy. Pulled the latest push and walked through each one:

1. **Shared `estimateCost`**: single code path now. The comment on the call site ("Use the same estimateCost() helper as goUsageTracker.record()") makes the intent obvious to the next reader.
2. **Tests**: 35 new cases in `goUsageTracker.test.ts`. Local run, 75/75 pass (was 40). Coverage hits accumulation, pruning, restoration, and the edge cases I flagged.
3. **`(est)` marker**: kept short, which I like. The SQLite schema argument is fair: no `modelId`/tokens/`sessionId` per row means per-session tracking has to live on the extension side. Wiring the reader for subscription-level totals is a separate concern (see follow-up below).
4. **Label**: `Latest Session (est)` reads better.
5. **Trailing newline**: fixed.

Compile clean. Tests green. Merging.

One real follow-up I'll file separately: wire the SQLite reader (`readOpenCodeHistory` / `buildSummaryFromRows`) into `getSummary()` so the subscription totals match what OpenCode actually bills. Right now those are dead code, and #23 was only worked around by PR #50's manual targets rather than fixed at the root. The cost drift between local estimate and server-side billing (cache affinity the extension can't see) is the underlying complaint there.

The threading of `sessionId` to the focused chat thread is blocked on a VS Code API that doesn't exist yet — `ProvideLanguageModelChatResponseOptions` exposes `requestInitiator` and `modelConfiguration` but no thread/session identifier. Leaving that one until the API lands.

Issue #58, opened 2026-06-25
[BUG] Agents window model picker


hu3bi:
The Agents window model picker is still bugged. E.g., showing models disabled by the user.

This applies to session local and copilot.

<img width="1086" height="733" alt="Image" src="https://github.com/user-attachments/assets/858c8323-28f0-44ff-9c98-f5abd89f3f8d" />

---

Model picker for copilot does not allow pinning or shows the cogwheel to display other models.

<img width="598" height="733" alt="Image" src="https://github.com/user-attachments/assets/b5cd80c1-9e22-4c7c-b539-18a8c87c2ebf" />

ltmoerdani:
Hey @hu3bi, thanks for flagging this. Reproduced on my end, and it's actually two separate things:

**1. Disabled models still showing in the Agents window picker**

This one is on us. OpenCode models for the Agents window are registered under separate vendor IDs (`opencodego-agent`, `opencodezen-agent`) to keep the regular Chat view clean. Problem: the enable/disable toggle in the *Manage Language Models* panel is per-vendor, and the agent vendors are hidden from that panel by default. So when you disable a model under `opencodego`, the matching `opencodego-agent` entry stays on and keeps showing up in the Agents window.

Fix in progress: syncing the enabled/disabled state across base and agent vendors, so disabling once covers both.

**2. No pin / cogwheel in the Copilot CLI picker**

This one's a VS Code limitation, not something we can fix from the extension side. The pin and cogwheel UI in the Copilot CLI model picker belongs to the Copilot extension itself, and there's no API for third-party language model providers to hook into it. Models registered with `targetChatSessionType: "copilotcli"` show up as regular dropdown options, but you can't pin them. For now the model has to be picked manually each session.

To make sure I'm fixing #1 against the same setup you're seeing, could you drop:
- VS Code version (from *Code → About*)
- Whether `extensions.supportAgentsWindow` has `ltmoerdani.opencode-copilot-chat` enabled
- A quick screenshot of your *Manage Language Models* panel, with the disabled models visible

That tells me which models you expected hidden vs. which actually appeared.

hu3bi:
To make sure I'm fixing https://github.com/ltmoerdani/opencode-copilot-chat/pull/1 against the same setup you're seeing, could you drop:

>    VS Code version (from Code → About):
Version: 1.126.0
Commit: 7e7950df89d055b5a378379db9ee14290772148a
Date: 2026-06-23T01:46:13Z
[...]

PR #60, opened 2026-06-29
feat: SQLite-backed cost accuracy, fix DeepSeek context overflow

### LINKED ISSUE #59: Wire SQLite reader into getSummary() to fix usage accuracy drift (#23 root cause)
## Context

PR #55 shipped session-level cost tracking on top of the extension's local estimate (`buildSummaryFromTracked`). That estimate is fine for per-session display. The **subscription totals** in the status bar tooltip (today / weekly / monthly) still use it, though, and the number can drift from what OpenCode actually bills. Two things the extension can't see: server-side cache affinity, and any usage from the CLI or TUI.

Issue #23 reported exactly this drift (weekly 9% to 15% too low, monthly 15% too low). It got closed as COMPLETED through PR #50, but PR #50 only added **manual usage targets**. The underlying estimate never changed.

## Current state (verified on `main` post-PR #55)

`goUsageTracker.ts` already has most of the pieces. They just aren't wired in:

- `readOpenCodeHistory()` (line 228) reads `~/.local/share/opencode/opencode.db`, returns `HistoryRow[]` with `{ createdMs, cost }`. SQL matches the OpenUsage plugin.
- `buildSummaryFromRows()` (line 343) builds a `UsageSummary` from those rows using the correct anchored-month formula.
- `hasSQLiteData` (line 424) is the only caller of `readOpenCodeHistory`, and itself has no callers.
- `getSummary()` (line 337) only calls `buildSummaryFromTracked(...)`. Never the SQLite path.

## Proposal

1. **Make SQLite the primary source.** `getSummary()` calls `readOpenCodeHistory()` first; if rows come back, use `buildSummaryFromRows()`. Fall back to tracked only when the DB is missing (user never ran the CLI).
2. **Merge tracked entries as enrichment**, with dedup by timestamp window so the two sources don't double-count.
3. **Adopt `anchoredMonthBounds`** from OpenUsage so the monthly window lines up with the dashboard. Current code uses calendar month, which accounts for part of the 15% drift on its own.
4. **Drop `hasSQLiteData`** unless it earns a real caller during the wire-up. Keep the code path if someone wants it as a feature flag, otherwise delete.

## Why this matters

Users treat the status bar as the source of truth for spend. A 9% to 15% drift, especially near the subscription cap, quietly trains them to distrust it. Manual targets move that burden onto the user; the default should be correct on its own.

The reader, SQL, and `buildSummaryFromRows` already exist. What's left is the call site swap, dedup logic, and the anchored month formula.

## References

- Issue #23 (closed COMPLETED via PR #50 workaround)
- PR #50 (manual targets, not a root fix)
- PR #55 (session tracking; unrelated to subscription totals, but it introduced the `(est)` marker pattern worth reusing here)
- OpenUsage plugin: https://github.com/robinebers/openusage/blob/main/docs/providers/opencode-go.md
- Repo memory note: `opencode-go-usage-accuracy.md` (TODO #1-#4, Jun 2026)

Wallacy:
### Goal

Fix subscription cost drift (9–15% vs actual billing, #59/#23) and prevent DeepSeek 400 errors when prompts are large.

### What was implemented

**SQLite-backed subscription costs (#59).** The extension tracked usage locally with `estimateCost()`, which can't see server-side cache affinity or CLI/TUI usage. The OpenCode CLI writes actual billed amounts to `~/.local/share/opencode/opencode.db`. Both `readOpenCodeHistory()` and `buildSummaryFromRows()` already existed but were dead code — `getSummary()` never called them. Now it tries SQLite first:

1. Reads `opencode.db` via the existing reader
2. Aggregates actual billed costs into the subscription buckets (session 5h, weekly, monthly, today, yesterday)
3. Enriches today/yesterday with token/request counts from tracked entries (SQLite doesn't store tokens)
4. Applies baselines on top of SQLite costs
5. Falls back to the local estimate when no SQLite DB exists — no breaking change

**DeepSeek context overflow fix.** When the prompt is large (e.g. 668K tokens on deepseek-v4-flash), `max_tokens` (384K) combined with the prompt exceeded the 1048K context window, causing a 400 error. `modelLimits()` now caps `maxOutputTokens` to `contextWindow - promptReserve` using the estimated prompt size. When the actual prompt token count is unknown, a conservative 80% reserve is used.

**SSE log cleanup.** `[sse-stats]` gated behind `debugReasoning` to reduce Output channel noise.

### Why not a web scraper / OAuth

The SQLite reader only works when the OpenCode CLI is installed (Linux/Mac, `~/.local/share/opencode/opencode.db`). Without it, the extension falls back to the local estimate (same as before). A proper OAuth or web-scraping approach would be the real long-term fix, but requires a public API or authentication flow that doesn't exist yet.

### Changelog

```
Added:
- SQLite-backed cost accuracy for subscription totals

Fixed:
- DeepSeek 400 error when prompt + completion exceeds context window
- [sse-stats] log gated behind debugReasoning
```

### Testing

- [x] Compiles, 75/75 unit tests pass
- [x] SQLite path: reads opencode.db, enriches tokens, applies baselines
- [x] No SQLite: falls back to tracked estimate (no regression)
- [x] DeepSeek: max_tokens now capped to prevent context overflow
- [x] SSE: [sse-stats] only appears with debugReasoning


ltmoerdani:
@Wallacy nice one, thanks for tackling both the cost drift and the DeepSeek overflow in one go.

The SQLite wiring is the part I find most useful. Right now the usage tracker guesses costs locally, which drifts 9-15% from what people actually get billed. Pulling real numbers from `opencode.db` fixes that trust gap, and the fallback when the DB is not there keeps it safe for people running without the CLI. Good call enriching tokens from tracked entries too, since SQLite only has cost data.

One FYI: `hasSQLiteData` getter and `getSummary()` both call `readOpenCodeHistory()` independently. If they fire close together that is two sqlite3 spawns for the same read. Probably fine for now, but if you ever see a hiccup in the status bar update, that is where I would look first.

The DeepSeek context cap is solid. `JSON.stringify` on the messages array will over-estimate token count a bit (JSON structure overhead), so the output budget ends up slightly tighter than the real limit. For safety that is actually the right direction, just wanted to flag it.

SSE gating, clean. That line showing up on every response was pure noise.

The commit history has that revert + re-apply of #58, net zero diff. If it was a merge conflict thing, maybe add a line in the PR description so people reading git log later do not wonder what happened.

Overall looks good to merge. Let me know if you want me to hit the button or if there is anything you want to tweak first.

Wallacy:
I just removed the commits rewriting the history. Now should be all okay.

ltmoerdani:
Looks clean now. 4 commits, no noise. Merge ready

Issue #67, opened 2026-07-01
[BUG] Annoying output window popup randomly


divyeshio:
## 🐛 What happened?

Output window pops up randomly, even though I dismiss it several times.

## 🔄 Steps to reproduce

1. Just send a request

## ✅ What did you expect?

I expect the window to stay closed.

## 🖥️ Environment

- **Extension version:** 0.3.4
- **VS Code version:** 1.127.0
- **Provider:** all
- **Model:** all

## 📸 Logs or screenshots (optional)

I believe this is happening due to these lines - 

https://github.com/ltmoerdani/opencode-copilot-chat/blob/7b20a2060002a41e541d5e08558c67eb000fd78b/src/extension.ts#L1280

https://github.com/ltmoerdani/opencode-copilot-chat/blob/7b20a2060002a41e541d5e08558c67eb000fd78b/src/extension.ts#L1291

Wallacy:
That's explain a lot... I was thinking that i was crazy and not remembering to change to the output.

ltmoerdani:
fyi this is fixed in PR #68, already merged to main. we removed all 3 `getOutputChannel().show(true)` calls that were causing the random popups. the output channel still works fine, it just won't force itself open anymore.

thanks @Wallacy for reporting this and fixing it in the same PR 🙏

closing this one since it's resolved.

divyeshio:
Thank you @Wallacy  @ltmoerdani  for the work you've been doing. When will this fix be released?

ltmoerdani:
@divyeshio , this shipped in v0.3.5 already! update the extension and the random popup should be gone. let us know if it comes back though 🙏

divyeshio:
@ltmoerdani  There are still few cases when it pops up for example stopping an ongoing chat, or sending stop and send message immediately.

ltmoerdani:
@divyeshio are you on v0.3.7? The popup fix went out in v0.3.5, so if you updated recently you should already have it.

If the popup still shows up when you hit stop or stop+send quickly, I need three things:

1. The version number from your Extensions panel (just to confirm)
2. Which model you had selected
3. Whatever the OpenCode output channel logged. To check: View > Output, pick "OpenCode" from the dropdown, grab the last 5 to 10 lines.

That should tell me if it's the same bug or something new.

divyeshio:
1. Yes, I'm on v0.3.7
2. It happens with every model, the one I'm using is deepseek-v4-flash-free

[...]

Issue #62, opened 2026-07-01
[FEATURE] Display only the list of enabled models


jmontornes-emdep:
## ✨ The feature

In Opencode ZEN, the list of available models includes all supported models, although the ones disabled from opencode.ai

## 🎯 The problem

A lot of models are listed, some of them very obsolete, although disabled from OpenCode web page

## 💡 Proposed solution

When fetching the models add the api key in the header. In this way, only the enabled models are included in the response.

It can be achieved just by modifying the extension.ts->fetchModels function:

`  private async fetchModels(): Promise<string[]> {
    try {
      const apiKey = await this.context.secrets.get(SECRET_KEY);
      const headers: Record<string, string> = {};
      if (apiKey) {
        headers["Authorization"] = `Bearer ${apiKey}`;
      }
      const response = await fetch(this.definition.modelsUrl, { headers });`

<img width="1528" height="233" alt="Image" src="https://github.com/user-attachments/assets/455cb13d-3c1a-43e3-a7f0-fc080ff24979" />

ltmoerdani:
This is done in PR #68, already merged to main. `fetchModels()` now accepts the API key and sends it in the Authorization header, so the OpenCode API only returns enabled models.

nice suggestion by the way, the fix was exactly what you described 👍

closing this since it's implemented.

jmontornes-emdep:
Thanks a lot for fast response!

PR #68, opened 2026-07-02
fix: GLM thinking values, remove output popup, filter enabled models — fixes #61, #67, #62

### LINKED ISSUE #61: [BUG] setting "opencodego.thinking.glm" allows invalid values
## 🐛 What happened?

I can set:

```yml
"opencodego.thinking.glm": "on",
```

However, `GLM 5.2` only provides `Off,` `High,` `Max`

---

**Why does this happen?**
Earlier versions of `GLM` e.g., `GLM 5.1` provided/used different values for the `thinking` property.

---

How to fix (options):

1. don't fix
2. have a fallback for `On`. It could fallback to `High`
3. remove default `thinking` setting
4. if all future `GLM` models provide those new values, change the default values of `thinking`.

I actually would prefer 2 or 3.

If we choose 4. the values for this setting have to be done automatically by pulling them from the opencode.ai endpoint. Else it creates too much maintenance in the future.

Wallacy:
fix: GLM thinking values, remove output popup, filter enabled models — fixes #61, #67, #62

### Goal

Fix three reported bugs: GLM accepting invalid thinking values, unwanted output channel popup, and model list showing disabled/unavailable models.

### What was implemented

**#61 — GLM thinking allows invalid values**
- Remove `"on"` from `opencodego.thinking.glm` enum — GLM 5.2 only accepts `off`/`high`/`max` per models.dev
- `buildThinkingPayload` now sends `reasoning_effort` for GLM when the value is effort-based (high/max); only `"off"` maps to `thinking: { type: "disabled" }`
- `buildFamilyThinkingSchema` split into separate blocks for GLM and Kimi (were combined in a single `if`)
- Updated `validate-models.mts` to test GLM with `high`/`max` instead of `on`
- Updated `package.json` with correct enum and concise description

**#67 — Annoying output window popup**
- Removed 3 `this.getOutputChannel().show(true)` calls in `testConnection()` (success and error paths) and in `provideLanguageModelChatResponse()` error catch

**#62 — Display only enabled models**
- `fetchModels()` already sent `Authorization: Bearer <key>` header, but the key was only read from secret storage internally. Now the API key is passed explicitly from `provideLanguageModelChatInformation()` to `fetchModels()`, ensuring the header always carries the correct key

### Changelog

```
Fixed:
- GLM 5.2 now only accepts off/high/max (removed invalid "on")
- GLM high/max now send reasoning_effort instead of thinking toggle
- Output channel no longer pops up randomly during requests
- Model list fetch always sends Authorization header with the correct API key
```

### Testing

86/86 unit tests pass. GLM dry-run on validate-models.mts confirms:
- GLM 5.1: `high` → `{"thinking":{"type":"enabled"}}`
- GLM 5.2: `high` → `{"reasoning_effort":"high"}`, `max` → `{"reasoning_effort":"max"}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,9 +172,9 @@
         },
         "opencodego.thinking.glm": {
           "type": "string",
-          "enum": ["on", "off"],
+          "enum": ["off", "high", "max"],
           "default": "off",
-          "markdownDescription": "Thinking mode for GLM models (`glm-5`, `glm-5.1`). Maps to `thinking: { type: 'enabled' | 'disabled' }`."
+          "markdownDescription": "Thinking mode for GLM models. `high`/`max` enable thinking (interpreted as `on` for models that use toggle, e.g., GLM 5/5.1). Per-model picker shows only options supported by each model."
         },
         "opencodego.thinking.kimi": {
           "type": "string",
--- a/scripts/validate-models.mts
+++ b/scripts/validate-models.mts
@@ -161,9 +161,10 @@ function buildThinkingTests(model: ModelInfo): ParamTest[] {
     }
     tests.push({ name: "no-reasoning (off)", settings: { deepseek: "off" } });
   }
-  // GLM: thinking type
+  // GLM: thinking type with effort values (GLM 5.2: high/max, older: enabled)
   else if (/^glm-/i.test(id)) {
-    tests.push({ name: "thinking=enabled", settings: { glm: "on" } });
+    tests.push({ name: "thinking=high", settings: { glm: "high" } });
+    tests.push({ name: "thinking=max", settings: { glm: "max" } });
     tests.push({ name: "thinking=disabled", settings: { glm: "off" } });
   }
   // Qwen: enable_thinking + budget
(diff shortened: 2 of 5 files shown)


ltmoerdani:
nice one, clean fix for 3 bugs at once. I went through the code:

**#61 GLM thinking** good approach sending `reasoning_effort` directly for effort-based values and letting the gateway handle resolution. Splitting `glm || kimi` into separate blocks makes sense too since they behave differently now. One small question: if a user already has `glm: "on"` saved in their workspace config, `buildThinkingPayload` will send `reasoning_effort: "on"` to the gateway. Will that fail gracefully or could it error out? If it might cause issues, might be worth adding a fallback `on` → `high` in `buildThinkingPayload` before it hits the gateway.

**#67 output popup** the `show(true)` calls were just in the wrong spots. Error handling via `showErrorMessage` is enough on its own.

**#62 filter models** clean, one-line change at the call site makes `fetchModels` more reusable. Nice.

102 lines of new tests covering effort values, schema generation, override, and invalid value rejection. Solid coverage.

I tested locally, works well. Merged ready. The stale GLM "on" config question is optional though, depends on how the gateway handles it, if it's graceful, just skip it.

Issue #93, opened 2026-08-02
OpenCode Go / Gpt 5.6 Luna not working


okerekechinweotito:
## 🐛 What happened?

OpenCode Go / Gpt 5.6 Luna not working 
always returns 
`Sorry, no response was returned.`

## 🔄 Steps to reproduce

1. select 5.6 luna from model picker
2. use agent mode and ask it to make an edit
3. errors out - Sorry, no response was returned.

## ✅ What did you expect?
should be able to make edits

## 🖥️ Environment

- **Extension version:**  
- latest
- **VS Code version:** 
- latest
- **Provider:** 
- OpenCode Go
- **Model:** 
- OpenCode Go / Gpt 5.6 Luna

ltmoerdani:
hey @okerekechinweotito, thanks for reporting this.

Turns out there were a few things going on:

1. GPT 5.6 Luna needs the Responses API endpoint (`/v1/responses`), not chat-completions. The extension was sending requests to the wrong endpoint entirely.

2. Tool calls in agent mode weren't showing up because the OpenCode Go gateway sends `finish_reason: null` on the final chunk instead of `"tool_calls"`. The extension was waiting for that signal to flush the accumulated tool calls, so they silently disappeared.

3. Thinking/reasoning wasn't working either. The reasoning payload wasn't being sent at all to the Responses API body, and the format was wrong (chat-completions style instead of Responses API style).

All three are fixed now. You should see it in the next release.

PR: https://github.com/ltmoerdani/opencode-copilot-chat/pull/97

okerekechinweotito:
@ltmoerdani 
Alright sounds good
Thanks for the fix

PR #100, opened 2026-08-03
fix(streaming): flush tool calls only at stream end (fixes #98)

### LINKED ISSUE #98: [BUG] VS Code Extension regressions cause malformed tool calls (<invoke> without <parameter>) in Chat sidebar
## 🐛 What happened?

In the latest version of the extension, using models such as DeepSeek-V4 via OpenCode Zen causes an unrecoverable tool-calling loop in the VS Code Chat panel.

The model attempts to invoke a tool but generates an incomplete or malformed payload—producing the opening `<invoke>` tag without any parameter elements or contents. The model then gets stuck attempting to manually type out parameters, but fails continuously.

Reverting to the previous version of the extension immediately resolves the issue.

*Note: The exact same DeepSeek model and OpenCode Zen provider work completely fine when run via the OpenCode CLI terminal client, isolating the issue strictly to the VS Code extension's system prompt injection or tool-parsing engine.*

## 🔄 Steps to reproduce

1. Update the **OpenCode** extension to the latest version.
2. Select `deepseek-v4` (or `deepseek-v4-pro`) using the **OpenCode Zen** provider in the VS Code Chat sidebar.
3. Send a prompt that requires tool execution (e.g., creating/updating a TODO list, reading/editing a file, or running a terminal command).
4. Observe that the model produces an empty `<invoke>` call missing `<parameter>` tags and enters an endless retry loop.

## ✅ What did you expect?

The model should output a fully structured tool call with populated parameters and execute the action:
```xml
<invoke name="tool_name">
  <parameter name="param_name">value</parameter>
</invoke>

xianhongtao:
The 0.4.4 #93 fix VS Code Extension regressions cause malformed tool calls (<invoke> without <parameter>) in Chat sidebar. 

## 📝 What does this change?

- OpenAiResponseExtractor now flushes only on finish_reason === "tool_calls"; remaining calls are flushed once at end-of-stream via the new flushRemainingToolCalls() (preserves #93 for gateways that omit finish_reason, e.g. gpt-5.6-luna on Go).
- Extracted a pure, unit-tested ToolCallAccumulator (src/toolCallAccumulator.ts, no vscode import) so the flush logic cannot silently regress again.
- Tests: src/test/toolCallAccumulator.test.ts (15 cases). All 121 unit tests + test-retry pass.
- Docs: CHANGELOG 0.4.5, docs/issues/42, devlog.

## 🧪 How did you test it?

I use deepseek via opencode go.

before fix:
image

after fix:
image

Note: gpt-5.6-luna (#93 path) not live-verified - China users cannot access GPT-series models via the gateway; covered by unit tests and the shared extractor code path, live check still recommended.

## ✅ Checklist

- [x] I tested it works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the **OpenCode Go BYOK Provider** extension are documented here.
 
+## [0.4.5] — 2026-08-03
+
+### Fixed
+
+- **`[Streaming]` Premature tool-call flush caused empty `<invoke>` calls and an unrecoverable tool-calling loop (#98).** The 0.4.4 fix for gpt-5.6-luna (#93) flushed accumulated tool calls whenever `finish_reason` was `null`/`undefined` and tool calls were pending. Standard OpenAI-compatible streams (DeepSeek-V4 via OpenCode Zen, plus Kimi, GLM, Qwen non-plus, MiniMax non-m2.x, MiMo) report `finish_reason: null` on EVERY intermediate chunk, so the first tool-call delta chunk flushed an INCOMPLETE call — empty arguments parsed as `{}`, and VS Code rendered `<invoke>` without `<parameter>`. The model then looped retrying the malformed call. `OpenAiResponseExtractor` now flushes ONLY on `finish_reason === "tool_calls"`; gateways that omit it (gpt-5.6-luna on OpenCode Go) are flushed once at end-of-stream via the new `flushRemainingToolCalls()`, so #93 stays fixed. Tool-call accumulation was extracted into a pure, unit-tested `ToolCallAccumulator` (`src/toolCallAccumulator.ts`, no `vscode` import). Documented in `docs/issues/42-20260803-premature-tool-call-flush.md`.
+
 ## [0.4.4] — 2026-07-25
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


ltmoerdani:
Thanks @xianhongtao for this. The root cause analysis checks out: the `finish_reason == null && pendingToolCalls.size > 0` condition in `streaming.ts` fires on basically every intermediate chunk, since most deltas carry `finish_reason: null`. Flushing there emits an incomplete call before arguments are accumulated, which is exactly what produces the empty `<invoke>` loop from #98. I confirmed it against current main.

Extracting `ToolCallAccumulator` into a pure module (no `vscode` import, mirroring `thinking.ts`) is the right move. Makes the flush logic unit-testable so this can't silently regress again, and the 15 test cases cover the important paths.

One thing before merging. The version is inconsistent across files.

- `CHANGELOG.md` says `0.4.5`
- `package-lock.json` bumps to `0.4.4`
- `package.json` isn't bumped at all (still on `0.4.4` from the #93 release)

Mind picking one version (either `0.4.5` in all three, or `0.4.4` if you'd rather fold this into the unreleased #93 bump) and syncing them? Happy to merge as soon as that lands.

Two minor things, not blockers:

1. `gpt-5.6-luna` (the #93 path) is only covered by unit tests here. The author notes China users can't reach GPT-series via the gateway, which is fair. I'll run a live check on my side after merge to make sure we didn't reintroduce #93. If anyone else can test sooner, even better.
2. Tiny naming nit: the extractor has `flushRemainingToolCalls()` while the accumulator has `flushRemaining()`. Not worth a rebase on its own, just flagging in case you revisit this code.

Appreciate the thorough write-up, and the honest note about the live-verification gap. Took the guesswork out of reviewing.

PR #102, opened 2026-08-03
fix(provider): improve Kimi context, model names, and image attachments

### LINKED ISSUE #87: [FEATURE] Ability to switch context size for Kimi models
## ✨ The feature

https://www.kimi.com/code/docs/en/kimi-code/models

Kimi modesl now has option for 256k context size, we should add ability to select them.

## 🎯 The problem

Current picker does not allow picking context size

### LINKED ISSUE #92: [FEATURE] Make provider prefixes in model names optional
## ✨ The feature

Add a setting that allows users to hide the `OpenCode Go /` and `OpenCode Zen /` prefixes from model names in the Copilot picker.

The setting could default to `true` to preserve the current behavior.

## 🎯 The problem

Models currently appear as, for example, `OpenCode Go / DeepSeek V4 Flash`.

However, Copilot already displays the provider separately for each row, so including it in the model name is redundant and makes the names harder to scan.

In a narrow chat view, the selected model may also be truncated, hiding the part that actually identifies the model. On a MacBook Air, I have to widen the chat pane to roughly one-third of the screen just to see the full model name.

## 💡 Proposed solution

Add a boolean setting, such as `opencodego.showProviderPrefix`, that controls whether the provider prefix is included.

When disabled, models would be displayed using only their model name.

## 🔄 Alternatives considered

- Keep the current prefixes for all users. This preserves the existing provider branding, but does not address truncation in narrow chat panes.

Making the prefixes configurable preserves the current behavior by default while allowing users to choose a more compact display.

## 📋 Additional context

This should be a display-only change. Model selection, session persistence, and request routing appear to use the model `id` rather than its display name.

Happy to open a PR if this would be useful.

### LINKED ISSUE #94: [BUG] Image attachments fail with HTTP 400 on OpenCode Go, while the same image+model works in opencode CLI
## 🐛 What happened?

Image attachments sent to vision-capable models via the OpenCode Go provider always fail with HTTP 400, even for small PNGs. The same image + same model works fine in the standalone opencode CLI.

Error (empty response body, `finish_reason: null` — the gateway rejects the request before the model runs):

```
OpenCode Go API request failed (400) model=gpt-5.6-luna payloadBytes=880950: {"id":"chatcmpl_pdsjmz976r","object":"chat.completion","created":1785660737,"model":"gpt-5.6-luna","choices":[{"index":0,"message":{"role":"assistant"},"finish_reason":null}]}
```

## 🔄 Steps to reproduce

1. Configure the OpenCode Go provider and select a vision-capable model (e.g. `gpt-5.6-luna`).
2. Attach any image (pasted screenshot or small PNG file) in the Copilot Chat input.
3. Send a message that includes the image.
4. The request fails immediately with the HTTP 400 error above.

## ✅ What did you expect?

The image should be sent to the model and answered, exactly as it works in the standalone opencode CLI (same model, same image).

## 🖥️ Environment

- **Extension version:** 0.4.3
- **VS Code version:** 1.128.0
- **Provider:** OpenCode Go
- **Model:** `gpt-5.6-luna`

Full error from the chat:

```
OpenCode Go API request failed (400) model=gpt-5.6-luna payloadBytes=880950: {"id":"chatcmpl_pdsjmz976r","object":"chat.completion","created":1785660737,"model":"gpt-5.6-luna","choices":[{"index":0,"message":{"role":"assistant"},"finish_reason":null}]}
```

Additional observations from reading the extension source (extension.js `convertMessage`):

- User-attached images (`LanguageModelDataPart` with `image/*` mime) are forwarded **as-is** as `image_url` with `data:<mime>;base64,...` — no resize, no re-encoding, no size limit. The only size cap (`MAX_TOOL_RESULT_IMAGE_BYTES = 1_000_000`) applies **only to tool-result images**, not user attachments.
- `payloadBytes=880950` confirms the image is actually sent (not stripped by the vision-proxy path), and the failure still occurs with small, spec-compliant PNGs.
- In contrast, opencode's client normalizes every image before sending (resize to ≤2000×2000, re-encode to PNG/JPEG, 5MB cap) via `image.normalize()`. The difference strongly suggests either (a) missing client-side image normalization, or (b) a request-body/endpoint contract mismatch with the `/zen/go/` gateway that only triggers when `image_url` content parts are present.

Wallacy:
## 📝 What does this change?

Implements fixes for #87, #92, and #94:

- Adds large-context options for Kimi/K3 models.
- Adds an optional provider prefix to model names.
- Normalizes oversized image attachments before sending them to OpenCode Go.

Closes #87
Closes #92
Closes #94

## 🧪 How did you test it?

- `npm test` passes: 132 tests.
- `npm run compile` passes.
- `npm run package` successfully generates the VSIX.
- Added automated coverage for model names, Kimi context options, and oversized image normalization.
- No live provider request was performed in VS Code.

## ✅ Checklist

- [x] I tested it works

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -2,6 +2,9 @@
 .vscode/**
 .vscode-test/**
 node_modules/**
+!node_modules/@silvia-odwyer/
+!node_modules/@silvia-odwyer/photon-node/
+!node_modules/@silvia-odwyer/photon-node/**
 src/**
 *.vsix
 tsconfig.json
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 All notable changes to the **OpenCode Go BYOK Provider** extension are documented here.
 
+## [Unreleased]
+
+### Added
+
+- **`[Model Picker]` Optional provider prefixes (#92).** Added `opencodego.showProviderPrefix` (default `true`) to hide `OpenCode Go` / `OpenCode Zen` prefixes in narrow model pickers when desired. Changes refresh the registered model names immediately.
+- **`[Model Picker]` Kimi context-size selector (#87).** Kimi models with a context window larger than 256K now expose `256K` and the full window in the per-model configuration schema, with the smaller tier selected by default.
+
+### Fixed
+
+- **`[Vision]` Normalize image attachments before OpenCode Go requests (#94).** Image data URLs are resized and re-encoded using the same 2000x2000 / 5 MB base64 limits used by OpenCode's CLI, while preserving the original image when normalization is unavailable.
+
 ## [0.4.5] — 2026-08-03
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -326,6 +326,7 @@ To manage agent API keys separately or see agent vendors in the Manage panel, en
 | `opencodego.requestTimeoutSeconds` | `600` | Total request timeout |
 | `opencodego.streamIdleTimeoutSeconds` | `120` | Cancel if stream goes idle |
 | `opencodego.showUsageStatusBar` | `true` | Show usage summary in status bar |
+| `opencodego.showProviderPrefix` | `true` | Include `OpenCode Go` / `OpenCode Zen` in model names |
 | `opencodego.freeOnly` | `true` | Zen: free models only. `false` = include paid |
 | `opencodego.agentsWindow` | `true` | Expose agent-host model variants (`targetChatSessionType`) for the Agents window |
 | `opencodego.showAgentModelsInManagePanel` | `false` | Show agent vendors in Manage Language Models panel |
--- a/docs/issues/44-20260803-issue94-image-normalization.md
+++ b/docs/issues/44-20260803-issue94-image-normalization.md
@@ -9,8 +9,10 @@ OpenAI `image_url` data URI, while the CLI normalizes images before sending.
 The request path now normalizes image data URLs with the same practical limits
 used by OpenCode: a maximum `2000x2000` image size and a `5 MB` base64 payload.
 It tries PNG first, then JPEG quality levels. If decoding or the optional
-normalizer fails, the original data URI is preserved so the provider behavior
-does not regress just because normalization is unavailable.
+normalizer fails, the original data URI is preserved and the final base64 guard
+decides whether it is still safe to send.
 
-The existing top-level and tool-result byte guards remain in place as a second
-line of defense against oversized conversation payloads.
+Top-level images are normalized before the final `5 MB` base64 guard, so an
+image larger than the old `2 MB` raw-byte threshold can still be sent when it
+can be reduced successfully. The separate `1 MB` raw-byte guard for tool
+results remains in place to bound cumulative MCP screenshot history.
--- a/docs/issues/44-20260803-issue94-image-normalization.md
+++ b/docs/issues/44-20260803-issue94-image-normalization.md
@@ -1,0 +1,16 @@
+**Status:** Implemented on `fix/open-issues-87-92-94`
+
+# Image Attachment Normalization (#94)
+
+OpenCode Go rejected some image attachments even though the same model and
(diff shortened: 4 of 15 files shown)


ltmoerdani:
Hey Wallacy, thanks for this. Took me a bit to go through all three fixes together.

The Kimi context tier (#87) and the provider prefix toggle (#92) look clean to me. `getContextSizeOptionsForModel` properly defers to models.dev tiers when they exist and only synthesizes the 256K option as a fallback, which is the right call. Extracting `formatModelName` into its own module is a nice cleanup too.

The image normalizer (#94) matches what opencode CLI does, so the 2000×2000 + 5MB thresholds are correct per the upstream contract. I checked the package, `@silvia-odwyer/photon-node` is WASM-based (~2.3MB, no native deps), which addresses the platform-specific concern I had from earlier image work. One thing I want to confirm before merging though.

There's a size guard in `convertMessage()` at line ~3453 that rejects top-level images above `MAX_TOP_LEVEL_IMAGE_BYTES = 2_000_000` raw bytes and replaces them with a placeholder text part. Your `normalizeImagePartsInPlace` runs *after* `convertMessage()` builds the messages, so in the current flow an image above 2MB raw gets dropped to a placeholder before the normalizer ever sees it. That means the normalizer only fires for images that are already under 2MB raw but somehow still trip the gateway (e.g. dimensions over 2000px on a small file).

Two questions on this:

1. Is that ordering intentional, or should `normalizeImageDataUrl` run *before* the size guard so we resize first and let the guard only catch images that are still oversized after normalization? My read of issue #94 (`payloadBytes=880950`, a sub-1MB PNG that still 400s) suggests the dimension-based normalization is the actual fix, and the existing guard is a separate safety net. But I want to make sure I'm not missing a case where both need to fire.

2. Related: should `MAX_BASE64_BYTES` (5MB) and `MAX_TOP_LEVEL_IMAGE_BYTES` (2MB raw) be reconciled, or are they meant to handle different layers? Right now they overlap in a slightly confusing way.

Also a minor one on `candidateSizes` in `imageNormalizer.ts`. The `reduce` with accumulator mutation and `.at(-1)` works, but it took me a couple of reads to follow the geometric decay logic. A plain `while` loop would read more obviously for what it's doing. Not blocking, just a readability nit.

No live provider test was run per your checklist, which is fair given the gateway behavior. I'm happy to test the normalizer path locally with a real vision model before we merge, or if you've already validated it works against the gateway, let me know and I'll take your word for it.

Tests look solid otherwise. Once we align on the guard ordering question, I'm good to merge this.

Wallacy:
Thanks for the careful review. You were right: the ordering was not intentional. The 2 MB raw top-level guard ran inside convertMessage() before image normalization, so it could replace images that Photon could have resized or compressed into a provider-safe payload.

I applied a follow-up that:

- normalizes top-level images before the final payload guard;
- removes the old 2 MB raw top-level guard;
- uses the shared 2000×2000 / 5 MB base64 limits as the final top-level contract;
- keeps the separate 1 MB raw guard for tool-result images, since that guard is intended to bound cumulative MCP screenshot history;
- rewrites candidateSizes() as a plain while loop;
- adds a regression test for a raw image above 2 MB whose base64 payload still fits the final limit.

Follow-up commit: `4572a9f`

ltmoerdani:
Yep, that nails it. I went through the diff again and the ordering is right now: `normalizeImagePart` fires before the guard on both paths (top-level around 3468, tool-result around 3405), and `convertMessage` being async doesn't break anything else since there's only the one caller at 2049.

Dropping `MAX_TOP_LEVEL_IMAGE_BYTES` and sharing `MAX_IMAGE_BASE64_BYTES` is the cleaner shape. One threshold to reason about, and the 1MB raw tool-result guard still doing its job for the MCP screenshot accumulation case from #38.

The while-loop rewrite and the 2MB-raw-but-5MB-base64 regression test, both good.

CI green on `4572a9f`, mergeable is clean. I'll do a merge commit to keep both your commits in history. Thanks

Issue #104, opened 2026-08-04
[BUG] DeepSeek 400 "Upstream response was not valid JSON" on 563KB session — payload trimmer (messageTrimmer.js) is dead code


aotr:
### Environment
- Extension: `opencode-copilot-chat` 0.4.5 (latest)
- Provider: OpenCode Go, model `deepseek-v4-flash` (also occurs across other models)
- VS Code: latest, macOS

### What happened
Sending a message in a long session fails with:

```
OpenCode Go API request failed (400) model=deepseek-v4-flash payloadBytes=563749: Error from provider (Console Go): Upstream request failed: [server_error] Upstream response was not valid JSON
```

### Steps to reproduce
1. Build a long session (~500KB+ of accumulated context / file attachments).
2. Send a message.

### Expected
The extension keeps the request body under the gateway's limit — it ships tooling for exactly this.

### Root cause
`out/messageTrimmer.js` defines `trimApiMessages()` with a 200KB messages budget and `MAX_PAYLOAD_BYTES = 380_000`, and its header states: *"The OpenCode Go API proxy returns HTTP 500 when the JSON request body exceeds ~400 KB."* But no code path calls `trimApiMessages` — a grep across `out/` shows it is referenced only inside `messageTrimmer.js` itself. The safety net is never wired up, so a 563KB body goes straight to the gateway and gets rejected.

#83 / PR #84 fixed context-window overflow by capping `maxOutputTokens` via `modelLimits()`, but that does not bound the raw request body, so this case still fails on the gateway body-size limit.

### Suggestion
Invoke `trimApiMessages()` with `MESSAGE_BYTE_BUDGET[endpointKind]` before `JSON.stringify(body)` in the request path (`out/streaming.js`), preserving the system prompt and tool-call atomicity as the module documents.

### Workaround
Starting a new chat session keeps the payload small and avoids the failure.

aotr:
Closing as withdrawn — investigation found the size hypothesis incorrect (gateway demonstrably accepts 1.5MB bodies, see issue #44). The failures were server-side on the gateway upstream.

Issue #83, opened 2026-07-24
[BUG]


gwynnbleiidd:
## 🐛 What happened?

<img width="1015" height="169" alt="Image" src="https://github.com/user-attachments/assets/b0941df3-3799-48ed-9559-8505ef476763" />

## 🔄 Steps to reproduce

1. Type message
2. Click send

## ✅ What did you expect?

Normal answer 

## 🖥️ Environment

- **Extension version:**  0.4.2
- **VS Code version:** 1.130.0 
- **Provider:** OpenCode Go
- **Model:** mimo-v2.5, deepseek-v4-flash and other

## 📸 Logs or screenshots (optional)

```
# OpenCode Go Diagnostics

## Recent Requests

### 1. hy3

- time: 2026-07-24T08:43:18.943Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 97122
- ttfbMs: 10556
- totalBytes: 0
- totalEvents: 0
- tokens: prompt=n/a, completion=n/a, total=n/a, cached=n/a
- cacheHitRatio: n/a
- finishReason: n/a
- requestId: req-8ca3ccb7
- sessionId: vscode-8249604f
- url: https://opencode.ai/zen/go/v1/chat/completions
- error: request cancelled

### 2. hy3

- time: 2026-07-24T08:41:35.820Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 100998
- ttfbMs: 10701
- totalBytes: 0
- totalEvents: 0
- tokens: prompt=n/a, completion=n/a, total=n/a, cached=n/a
- cacheHitRatio: n/a
- finishReason: n/a
- requestId: req-b0345245
- sessionId: vscode-18482757
- url: https://opencode.ai/zen/go/v1/chat/completions
- error: request cancelled

### 3. mimo-v2.5

- time: 2026-07-24T08:38:07.797Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 4029
- ttfbMs: 3974
- totalBytes: 976
- totalEvents: 5
- tokens: prompt=754773, completion=1, total=754774, cached=754752
- cacheHitRatio: 100.0%
- finishReason: length
- requestId: req-bc031363
- sessionId: vscode-086ceb80
[...]

Issue #44, opened 2026-06-15
timetable arrangement 


arpitbahetiap-jpg:
To do all the two tasks at a time in 3.5hrs with minimal screen time

PR #105, opened 2026-08-04
fix: wire up payload trimmer to fix DeepSeek 400 "Upstream response was not valid JSON" on large sessions (#104)

aotr:
### Environment
- Extension: `opencode-copilot-chat` 0.4.5 (latest)
- Provider: OpenCode Go, model `deepseek-v4-flash` (also occurs across other models)
- VS Code: latest, macOS

### What happened
Sending a message in a long session fails with:

```
OpenCode Go API request failed (400) model=deepseek-v4-flash payloadBytes=563749: Error from provider (Console Go): Upstream request failed: [server_error] Upstream response was not valid JSON
```

### Steps to reproduce
1. Build a long session (~500KB+ of accumulated context / file attachments).
2. Send a message.

### Expected
The extension keeps the request body under the gateway's limit — it ships tooling for exactly this.

### Root cause
`out/messageTrimmer.js` defines `trimApiMessages()` with a 200KB messages budget and `MAX_PAYLOAD_BYTES = 380_000`, and its header states: *"The OpenCode Go API proxy returns HTTP 500 when the JSON request body exceeds ~400 KB."* But no code path calls `trimApiMessages` — a grep across `out/` shows it is referenced only inside `messageTrimmer.js` itself. The safety net is never wired up, so a 563KB body goes straight to the gateway and gets rejected.

#83 / PR #84 fixed context-window overflow by capping `maxOutputTokens` via `modelLimits()`, but that does not bound the raw request body, so this case still fails on the gateway body-size limit.

### Suggestion
Invoke `trimApiMessages()` with `MESSAGE_BYTE_BUDGET[endpointKind]` before `JSON.stringify(body)` in the request path (`out/streaming.js`), preserving the system prompt and tool-call atomicity as the module documents.

### Workaround
Starting a new chat session keeps the payload small and avoids the failure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the **OpenCode Go BYOK Provider** extension are documented here.
 
+## [0.4.6] — unreleased
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


aotr:
Closing per the captain's decision — no longer pursuing this fix. Corrected finding: the gateway accepts ~1.5MB bodies (see issue #44), so the ~400KB size theory was wrong; failures were server-side.

PR #107, opened 2026-08-05
fix(streaming): retry transient 5xx gateway errors

Fahad090NP:
### 1. Transient 5xx retry (main fix)
- `src/retry.ts`: add `isTransientServerError(status, errorDetail)` — 502/503/504 are retried; other 5xx only when the body names `Router.Unavailable` (no healthy backend for the model). Add `TRANSIENT_5XX_MAX_RETRIES` / `TRANSIENT_5XX_RETRY_BASE_MS` constants. Header comment rewritten to document the two retry families (400 degraded-parameter + transient 5xx).
- `src/streaming.ts`: on a transient 5xx, retry up to 2 times with exponential backoff (1s → 2s), honoring cancellation between attempts (`sleepWithCancellation`); request fetch refactored into `fetchWithBody`; stale consumed error body reset after each retry. Also fixes 3 pre-existing lint issues in the file (unused var, `let`→`const`, unused eslint-disable).
- `src/errors.ts`: when the gateway reports `Router.Unavailable`, show an actionable hint ("no healthy backend for this model right now — retry in a few seconds or switch models") instead of raw JSON.
- `src/test/retry.test.ts`: unit tests for `isTransientServerError`.

### 2. Developer tooling
- `.husky/pre-commit`: run `lint-staged` on commit.
- `package.json`: `prepare: husky`, `lint-staged` config (eslint `--fix --max-warnings 0` + prettier on js/ts; prettier on json/yml; prettier + markdownlint on md), prettier config (`trailingComma: none`), new scripts (`lint`, `lint:js`, `lint:md`, `format`, `format:check`).
- `eslint.config.mjs` (new): flat config, `typescript-eslint` recommended, honoring `.gitignore`.
- `.markdownlint.json` (new): disable MD013.
- `package-lock.json`: entries for new devDependencies.

### 3. Build
- `tsconfig.json`: add `"include": ["src"]` so tsc only compiles source files.

### 4. Docs
- `AGENTS.md` (new): contributor/agent instructions (philosophy, workflow, verification, architecture, style, error handling).

## How did you test it?
- `npm run compile` — clean.
- `npm test` — 138/138 pass.
- `npm run test-retry` — E2E retry flow, 7/7 pass.
- `eslint --max-warnings 0` on changed files — clean.

## Checklist
- [x] Changes tested
- [x] Docs updated (AGENTS.md); CHANGELOG untouched (release-time only)

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -60,11 +60,25 @@ export function buildOpenCodeRequestError(
     );
   }
 
-  const userMessage = `${providerDisplayName} API request failed (HTTP ${response.status})${modelHint ? ` for ${modelId}` : ""}: ${apiMessage}${capacityHint}`;
-  return new OpenCodeRequestError(
-    `${providerDisplayName} API request failed (${response.status})${modelHint}${sizeHint}${capacityHint}: ${apiMessage}`,
-    userMessage,
-  );
+  const userMessage = `${providerDisplayName} API request failed (HTTP ${response.status})${modelHint ? ` for ${modelId}` : ""}: ${describeRouterUnavailable(apiError, apiMessage)}${capacityHint}`;
+  const requestMessage = `${providerDisplayName} API request failed (${response.status})${modelHint}${sizeHint}${capacityHint}: ${apiMessage}`;
+  return new OpenCodeRequestError(requestMessage, userMessage);
+}
+
+/**
+ * Replace the raw gateway JSON detail with an actionable hint when the error
+ * is the transient `Router.Unavailable` condition (no healthy backend for
+ * the model right now). Returns the original message otherwise.
+ */
+function describeRouterUnavailable(
+  apiError: ParsedApiError,
+  fallback: string,
+): string {
+  const type = apiError.type?.toLowerCase() ?? "";
+  if (!type.includes("routerunavailable")) {
+    return fallback;
+  }
+  return "OpenCode's router has no healthy backend for this model right now. Retry in a few seconds, or switch to another model.";
 }
 
 export function truncateForLog(value: string, max = 1200): string {
--- a/src/retry.ts
+++ b/src/retry.ts
@@ -1,18 +1,29 @@
 /**
- * Runtime retry with degraded parameters for HTTP 400 errors.
+ * Runtime retry decisions for HTTP failures.
  *
- * When the upstream API rejects a parameter (e.g. thinking.type "disabled",
- * invalid temperature, unsupported reasoning_effort), this module:
- * 1. Parses the error message to identify the problematic parameter
- * 2. Removes or adjusts it in the request body
- * 3. Returns the patched body for a single retry
+ * Two distinct retry families are handled here:
+ * 1. Degraded-parameter retry for HTTP 400 errors (see analyzeHttp400ForRetry).
+ * 2. Transient 5xx classification (see isTransientServerError) for short
+ *    backoff retries when the gateway router is temporarily unavailable.
  *
  * CONTRACT:
  * - Pure functions only — no vscode import, no side effects.
- * - Only handles recoverable 400 errors. Auth (401/403), rate limit (429),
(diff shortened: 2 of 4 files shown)


ltmoerdani:
nice one @Fahad090NP the transient 5xx retry is solid logic. `isTransientServerError` keeping it conservative (502/503/504 + `Router.Unavailable` only, not blanket 5xx) is the right call, and `sleepWithCancellation` registering the listener before the timer is a subtle detail I appreciate.

a few things before I can merge though.

**scope split.** the PR title says `fix(streaming)` but 6 of 11 files are tooling/docs/build (husky, lint-staged, eslint flat config, markdownlint, AGENTS.md, tsconfig tweak). the core retry fix in `retry.ts` / `errors.ts` / `streaming.ts` / `retry.test.ts` is clean and I want to land it fast. can we split this into two PRs?

- **PR 1 (this one, narrowed):** just the retry fix + tests. low-risk, high-value, I can merge quickly.
- **PR 2 (separate):** the husky + eslint + AGENTS.md stack. this changes the workflow for every future contributor, so I want to think about it separately and run `npm run lint` against the current baseline first to make sure we're not gonna block commits on day one.

**one thing on the backoff itself:**

```ts
const backoffMs = TRANSIENT_5XX_RETRY_BASE_MS * 2 ** (attempt - 1); // 1s → 2s
```

no jitter. if a batch of concurrent requests (multi-turn agent, parallel tool calls) all eat a 502 in the same window, they'll all retry at the exact same timestamps and pile on the gateway. mind adding a small `+ Math.random() * 250` or full jitter? cheap fix, real upside under load.

**quick question on the 400→5xx handoff:** when the first request 400s, gets patched, then the patched retry returns 503, we now enter the 5xx loop on top of the 400 retry. that's up to 4 fetches for one user request. not wrong, just want to confirm it's intentional and that resetting `consumedErrorBody = undefined` at the end of the 400 block is meant so the 5xx block reads a fresh body. sanity check passes for me, but worth a yes/no.

**on AGENTS.md:** it's well-written, but a couple of lines conflict with how I run the project. e.g. "Never run dev server or build to check for errors, that's the user's job" vs. my workflow where I expect `npm run compile` before claiming a task is done. also "Never auto-commit" is already covered in CONTRIBUTING.md. rather than ship a competing doc, let's fold the parts you care about into CONTRIBUTING.md in PR 2 and skip a standalone AGENTS.md. open to being wrong here if you feel strongly.

tests look good. once the split happens and jitter lands, I'm happy to merge PR 1 fast.

Fahad090NP:
Thanks for the feedback — all points addressed on this branch:

- **Scope split** — Done. `feat/error-handling` is narrowed to the retry fix only (`errors.ts`, `retry.ts`, `streaming.ts`, `retry.test.ts`). The husky/eslint/markdownlint/tsconfig/CONTRIBUTING stack moved to its own PR (#110).
- **Jitter** — Added. Backoff is now `TRANSIENT_5XX_RETRY_BASE_MS * 2 ** (attempt - 1) + Math.random() * TRANSIENT_5XX_RETRY_JITTER_MS` (≤250ms), so concurrent retries spread out instead of piling on the gateway.
- **400→5xx handoff** — Confirmed intentional: `consumedErrorBody` is reset to `undefined` at the end of the 400 block (and again per 5xx retry) so the 5xx block reads a fresh response body. Yes, up to 4 fetches for one user request is by design.
- **AGENTS.md** — Agreed, dropped the standalone doc; the relevant guidance is folded into CONTRIBUTING.md in #110.

Tests unchanged and passing. Happy for this to merge.

ltmoerdani:
Thanks for the clean split — PR #107 is in good shape. jitter landed correctly, and the 400→5xx handoff reasoning is clear now that you confirmed the reset is intentional.

I'll run `npm run compile`, `npm test`, and `npm run test-retry` on the branch locally to double-check, then merge. PR #110 I'll review separately.

Issue #109, opened 2026-08-06
[BUG] DeepSeek V4 Flash request exceeds context limit


TDC-TangChaoBanLi:
## 🐛 What happened?

When using deepseek-v4-flash through OpenCode Go, Copilot Chat fails with HTTP 400:

This model's maximum context length is 1048576 tokens.
However, you requested 1050237 tokens
(666237 in the messages, 384000 in the completion).

## 🔄 Steps to reproduce

1. Install extension version `0.5.0`.
2. Configure OpenCode Go.
3. Select `deepseek-v4-flash`.
4. Use GitHub Copilot Chat or Agent Mode with tools enabled.
5. Continue a conversation until the displayed context reaches approximately 650K.
6. Send another message.
7. Observe that the extension sends `max_tokens: 384000`.
8. The provider rejects the request because prompt tokens plus maximum completion tokens exceed 1,048,576.

The failure may depend on the number and size of the Copilot/MCP tool schemas available in the session.

## ✅ What did you expect?

--

## 🖥️ Environment

- **Extension version:**  0.5.0
- **VS Code version:** 1.132.0
- **Provider:** OpenCode Go 
- **Model:** deepseek-v4-flash

## 📸 Logs or screenshots (optional)

vscode copilot report :

<img width="969" height="2550" alt="Image" src="https://github.com/user-attachments/assets/f8f68785-f5c9-4619-a782-78e1462e9a17" />

```text
Sorry, your request failed. Please try again.

Client Request Id: 66901a26-8a4c-4e8c-bc96-1849c2a9a9dd

[...]

Issue #103, opened 2026-08-03
[BUG] gpt-5.6-luna: invalid_prompt (400) on long sessions via Responses API


ltmoerdani:
`gpt-5.6-luna` dies with HTTP 400 once a session crosses roughly 1.5 MB of payload. Short chats are fine.

```
OpenCode Go API request failed (400) model=gpt-5.6-luna payloadBytes=1528273:
Error from provider (Console Go): Upstream request failed:
[invalid_prompt] Invalid Responses API request
```

After this point every subsequent request in the same session fails. New session resets it.

## Steps to reproduce

1. Pick `gpt-5.6-luna` in Copilot Chat (OpenCode Go provider)
2. Start a fresh session in agent mode with tools
3. Keep going: tool outputs, file reads, code blocks, the usual agent loop
4. Around 1.5 MB payload, every request starts returning `invalid_prompt`

## What I found

`buildResponsesRequestBody()` in `src/extension.ts` has three gaps. I went in suspecting the gateway, came out blaming us.

**1. `truncation` is never sent. Defaults to `disabled`.**

Per the OpenAI Responses API docs, `truncation: "disabled"` (the default) means the request fails with 400 once input exceeds the context window. We never set `"auto"`. Long sessions hard-fail instead of truncating. This is almost certainly the main cause.

**2. `max_output_tokens` is not capped against the remaining window.**

`gpt-5.6-luna`: context window 1,050,000, `max_output_tokens` 128,000. The request fails the moment the prompt exceeds 922,000 tokens. We ship the static number from metadata without subtracting input size.

**3. `text.verbosity` is sent on every Responses call.**

Native OpenAI field. But the OpenCode Go gateway is a proxy in front of other providers, and there's no guarantee they forward it. Providers that reject unknown fields would return `invalid_prompt` from this alone.

## Why this is not a duplicate of #93 / #41

#93 and #41 fixed response parsing: the gateway sends `finish_reason: null` and our extractor didn't flush tool calls. Those work now.

This is one layer earlier. The request never reaches the response parser because the provider rejects it up front.

## Proposed fix

1. Send `truncation: "auto"` on the Responses body. This one alone probably fixes the bulk of long-session failures.
2. Cap `max_output_tokens` against `contextWindow − estimatedPromptTokens`. Handles the edge case.
3. Drop `text.verbosity`. Lower priority, but free.

[...]

Issue #89, opened 2026-07-31
[FEATURE] VS Code extensions are incompatible when running VS Code as administrator


dqd1505826212:
## ✨ The feature

When launching Visual Studio Code with administrator privileges, its extensions fail to work and show compatibility issues.

ltmoerdani:
Thanks for reporting, and sorry for the slow reply. We got partway there in #113: the extension now captures runtime diagnostics specifically for elevated/admin launches, so we can see which VS Code binary got selected, where the extension installed, workspace trust state, and Windows integrity level.

To pin down a real fix, could you run `OpenCode Go: Diagnostics` (or `OpenCode Zen: Diagnostics`) inside the elevated VS Code where the issue appears, then paste the Runtime section plus a failing request summary here? The admin case usually boils down to VS Code picking a different install path or SecretStorage being locked, and that output tells us which one.

Closing for now, one diagnostic paste away from a proper fix.

Issue #88, opened 2026-07-30
[FEATURE] OpenCode server local models support ?


nvandamme:
I've got a local instance of opencode server running with additional providers to run a truly local setup (opencode-lmstudio@0.3.1). Could it be possible to also expose local opencode server model list to vscode copilot ?

ltmoerdani:
Thanks for the request, and sorry for the slow reply. We looked at this properly during #113, and the short version is: it's a technical blocker for now.

`opencode serve` exposes OpenCode sessions/agents, not a Chat Completions, Responses, or Messages endpoint that can return tool calls to VS Code. Registering only the model list would break Copilot's tool loop and permission UI, so we documented the decision in the README FAQ instead of shipping half-working entries.

For a local setup that works today, point VS Code's Custom Endpoint provider at any OpenAI/Anthropic-compatible server (Ollama, LM Studio, llama.cpp, etc.). Full tool loop, local models.

Closing as not planned for now. Happy to revisit if the provider contract changes.

PR #113, opened 2026-08-07
feat: harden OpenCode VS Code bridge

Wallacy:
## Summary

Harden the OpenCode language-model bridge against current VS Code provider behavior, long-running Responses API conversations, tool-heavy agent sessions, and Extension Host cold starts.

## Changes

- Fix context-budget failures in long and tool-heavy sessions (#103, #109):
  - Enable `truncation: "auto"` for Responses API requests.
  - Remove the proxy-sensitive `text.verbosity` field.
  - Include Copilot/MCP tool definitions and JSON schemas in prompt estimates.
  - Reserve proportional tokenizer headroom instead of a fixed 64-token margin.
  - Bound the requested output to the context remaining after the normalized prompt.
  - Recover from exact upstream HTTP 400 context overflows by using the provider-reported counts to reduce the output budget and retry once.
  - Cover Chat Completions, Anthropic Messages, Responses, and Gemini output-budget formats with pure regression tests.
- Improve current VS Code integration:
  - Mark models as BYOK and expose capacity warnings.
  - Add provider management commands and explicit utility-model configuration.
  - Keep vision and tool calling enabled without advertising `capabilities.editTools`, which is still gated behind the `chatProvider` proposed API in VS Code 1.132. This field is only an edit-tool preference hint; omitting it does not disable agent edits or tool calls.
- Restore credentials for cached models after an Extension Host restart by falling back to SecretStorage before reporting a missing API key.
- Add runtime diagnostics for VS Code version, host, remote mode, workspace trust, Windows elevation, paths, credentials, and model-selection failures.
- Fix cancellation listener leaks and retry timeout handling.
- Clean `out/` before compilation to prevent stale artifacts from entering the VSIX.
- Make linting, tests, and packaging blocking CI checks.
- Reduce development-only files shipped in the VSIX.
- Update README, changelog, issue notes, and contributor checklist.

## Scope Notes

- `opencode serve` is documented but not registered as a native VS Code model source because its session/agent API is not a compatible inference endpoint for VS Code's tool loop. See [OpenCode server documentation](https://opencode.ai/docs/server/) and #88.
- Inline autocomplete remains dependent on upstream VS Code/Copilot support.
- #89 now has richer diagnostics, but a reproducible elevated/admin failure is still needed for a definitive fix.
- The #109 failure is now covered structurally and by the exact reported token counts; final validation with the reporter's near-limit live session is still pending.
- Proposal-only edit-tool preferences remain intentionally disabled for Marketplace compatibility and can be restored when `chatProvider.editTools` becomes stable.

## Verification

- `npm run compile`
- `npm run lint`
- `npm test` - 156 tests passing on Windows
- `npm audit --omit=dev` - 0 vulnerabilities
- `npm run package` - clean 39-file VSIX generated successfully
- Installed the branch VSIX in VS Code 1.132; Manage Models lists the OpenCode models without a `chatProvider` proposal error.
- Verified the packaged capability object keeps vision/tool calling enabled and does not contain `editTools`.
- Sent a live request through VS Code with OpenCode Go / Deepseek V4 Flash and received the expected response.
- Restarted the Extension Host and sent another request with the cached selected model, without reopening Manage Models; the request completed through the provider instead of reporting a missing API key.

Live gateway smoke tests now pass. Only a full near-limit production session for #109 remains pending.

Addresses #109.


ltmoerdani:
Nice one on this @Wallacy pulled the branch locally and ran the usual checks, all green:
- `npm run compile` clean
- `npm test` 156/156 passing
- `npm run lint:js` and `lint:md` clean

The context hardening is solid. Counting tool schemas in the estimate, swapping the fixed 64-token margin for a proportional one, and trusting the upstream counts on the 400 recovery all stack up into a proper safety net. I ran the math on the #109 numbers myself, request lands under the 1,048,576 limit. Also good call extracting the responses envelope, limits, and token estimate into pure modules with tests. Those would've been rough to eyeball buried inside `extension.ts`.

Three small things, none blocking:

1. `provideLanguageModelChatResponse` now awaits `secrets.get(SECRET_KEY)` every request to cover the cold-start case. Makes sense, VS Code caches SecretStorage reads so the overhead should be minor. Just want to confirm it isn't adding visible latency on slower machines, or if a short-lived cache would be safer.
2. `truncation: "auto"` on the Responses body can drop earlier turns, system message included, once it kicks in. No user-facing notice when that happens. The issue doc covers the trade-off so it's documented. Still, curious if you considered surfacing something when truncation actually triggers.
3. `lint:md` now skips `docs/**`. Makes sense since docs are excluded from the VSIX, but CI won't catch formatting slips in the issue/feature notes anymore. `lint:md:all` exists, so maybe worth running it as a non-blocking job.

One note before merge: I'll merge with a regular merge commit to keep all four commits, same as your previous PRs.

PR #120, opened 2026-08-08
feat(vision): cache vision-proxy image descriptions

### LINKED ISSUE #119: [Enhancement] Vision proxy re-describes images on every turn - no cache
## Problem

When using a text-only model with the vision proxy enabled, `proxyVision()` calls the vision model again on **every** turn to describe the same images. This:

- Wastes token
- Makes each turn slower
- Can give different descriptions each time

## Suggestion

Cache the description by image hash (like the code already does for `reasoningContentByToolCallId`). Reuse it when the image has not changed. Only call the vision model for new images.

## Environment

- Extension: latest
- Setup: text-only model + vision proxy enabled

ChauThan:
## Summary

Caches the text description produced by the vision proxy per image (SHA-256 of its base64 bytes), so images that were already described in earlier turns are reused without calling the vision model again. This saves Copilot quota and latency in multi-turn conversations where text-only models receive the same image attachments on every turn.

## What changed

- **New `src/visionProxyCache.ts`** - an in-memory cache keyed by image content hash, capped at 200 entries with FIFO eviction (mirroring the reasoning-content cache).
- **`src/extension.ts`** - `proxyVision()` now:
  - describes only the message that contains a **new** image (instead of re-sending the whole conversation), and
  - reuses cached descriptions when all images in a message were already described - no vision-model request at all in that case, so `selectChatModels()`/`sendRequest()` are skipped lazily.
  - messages are flattened before the vision pass so the correct description is applied to the right `apiMessage` when one input message expands into several (e.g. tool results).
- **`src/test/visionProxy.test.ts`** - unit tests for the cache (stable keys, lookup/store, partial hits, reuse, FIFO eviction, clear).
- **`package-lock.json`** - version synced to `0.5.1`.

## Why

Without a cache, `proxyVision()` calls the vision model via `model.sendRequest()` on every turn to describe the same bytes again - wasting Copilot quota, adding latency, and returning a different description each time.

## Validation

- `npm run compile` passes
- `npm test` passes - 168 tests (7 new for the cache)
- `npm run lint:js` passes

## Issue
Closes #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -339,6 +339,7 @@ Provider diagnostics also include the VS Code/extension versions, extension host
 | `opencodego.streamIdleTimeoutSeconds`     | `120`    | Cancel if stream goes idle                                                       |
 | `opencodego.showUsageStatusBar`           | `true`   | Show usage summary in status bar                                                 |
 | `opencodego.showProviderPrefix`           | `true`   | Include `OpenCode Go` / `OpenCode Zen` in model names                            |
+| `opencodego.visionProxyWholeConversation` | `false`  | Vision proxy: describe the whole conversation instead of only the message with a new image (more context, more tokens) |
 | `opencodego.freeOnly`                     | `true`   | Zen: free models only. `false` = include paid                                    |
 | `opencodego.agentsWindow`                 | `true`   | Expose agent-host model variants (`targetChatSessionType`) for the Agents window |
 | `opencodego.showAgentModelsInManagePanel` | `false`  | Show agent vendors in Manage Language Models panel                               |
--- a/docs/features/11-20260715-vision-proxy.md
+++ b/docs/features/11-20260715-vision-proxy.md
@@ -79,6 +79,20 @@ Storage keys (in extension `globalState`):
 - `opencodego.visionProxyModelId` — target Copilot model ID (e.g. `copilot:gpt-5.5`)
 - `opencodego.visionProxyPrompt` — description instruction sent to the vision model
 
+### Description Cache & Whole-Conversation Mode
+
+Descriptions produced by the proxy are cached per image (SHA-256 of its bytes), so
+images already described in earlier turns are reused without calling the vision model
+again - saving Copilot quota and latency in multi-turn conversations (issue #119).
+
+A boolean setting controls how much context each description gets:
+
+- `opencodego.visionProxyWholeConversation` (default `false`). Off: the proxy
+  describes only the message that contains a new image and reuses cached
+  descriptions, keeping token usage low. On: the proxy sends the whole
+  conversation to the vision model so descriptions carry full context, at the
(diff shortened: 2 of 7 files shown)


ltmoerdani:
Nice one @ChauThan this is a clean fix. the cache mirrors the reasoning-content cache nicely, and the 7 tests cover the important cases (the partial-hit one especially).

a couple of things i want to double check before merging:

1. multi-image messages. you store the combined description under every image hash, so if one of those images shows up alone in a later message it gets a description that also mentions the other image. not a correctness issue since a partial hit still forces a re-describe, just confirming that's intended.
 
2. the proxy now describes only the message that has a new image instead of the whole conversation. that changes the behavior a bit, descriptions lose some conversation context. makes sense for the token savings, just wanted to flag it.

also, the lockfile sync to 0.5.1 fixes a mismatch that was sitting on main (package.json was already 0.5.1 but the lockfile was still 0.5.0), so that's a welcome side effect. the extra `peer: true` flags on a few dev deps look like npm version noise, harmless either way.

i'll run compile + the full test suite locally before merging. thanks for the PR.

ChauThan:
@ltmoerdani 
Thanks for the review!

1. Yes, that's intentional. I store the same description under every image hash. A partial hit returns nothing, so the message gets described again - no risk of wrong descriptions.

2. That was not intentional - I didn't think it through carefully. I'm thinking about adding a setting flag so users can choose whether to accept this trade-off or not. Thanks for pointing it out.

I'll do it if you agree.

ltmoerdani:
@ChauThan thanks for the follow-up.

1. good, that matches what i read in the code. partial hit forces a re-describe, so no stale description gets reused.

2. sure, go ahead with the setting flag. a couple of guardrails so it stays small: keep the cache behavior as the default, gate the whole-conversation mode behind the flag, and add a short docs entry for it. ping me when it's ready and i'll review + test before merging.

ChauThan:
@ltmoerdani done - added the setting flag as discussed.

- `opencodego.visionProxyWholeConversation` (default `false`) - off keeps the cache behavior (describe only the message with a new image and reuse cached descriptions), on describes the whole conversation so descriptions keep full context (still cached, more tokens).
- `proxyVision()` now takes `describeWholeConversation`; the whole-conversation branch sends one request over all messages and stores the combined description under every image hash (same no-partial-reuse rule as before).
- Docs: README settings table + a short section in the vision-proxy feature doc.
- compile / test (168 pass) / lint:js all green.

Ready for review + test whenever. Thanks!

ltmoerdani:
@ChauThan nice one, this rounds it out well. i pulled the branch and ran the checks locally: compile passes, 168 tests pass (0 fail), lint:js and markdown lint on the feature doc are clean. the setting registration, README table, and the feature doc section all line up. merging now, thanks!

PR #124, opened 2026-08-10
fix(vscode): drop managementCommand from model providers to restore native BYOK group flow (#121)

Fahad090NP:
Closes #121

## Problem

The `languageModelChatProviders` contributions declared **both** `managementCommand` and a `configuration` schema. VS Code's built-in `configureLanguageModelsProviderGroup()` **short-circuits on `managementCommand`** — it just re-resolves the models and returns, never prompting for a group name or API key. Consequences:

- **"+ Add Models"** clicking *OpenCode Go* / *OpenCode Zen* did nothing (no API key prompt, no group created).
- **Context menu unresponsive** — *Rename Group*, *Delete*, *Update API Key*, *Open in Language Models (JSON)* all operate on groups in `chatLanguageModels.json`, but no group could ever be created through the UI, so every action throws `Language model provider group … not found` and fails silently.
- **Leftover default group can't be removed** — e.g. the `OpenCode Go` group persisted in `chatLanguageModels.json` (created by per-model `reasoningEffort` configuration) could never be deleted.

## Fix

Removed `managementCommand` from the `opencodego`, `opencodezen`, `opencodego-agent`, and `opencodezen-agent` contributions. With only the `configuration` schema present, VS Code's native BYOK flow runs end-to-end:

- `+ Add Models` prompts for a group name + API key and creates the group.
- Rename / Update API Key / Delete / Open in Language Models (JSON) now work against that group.
- The stale/leftover group can finally be deleted.

The extension's own management commands (`OpenCode Go: Manage Provider`, `OpenCode Go: Set API Key`, …) remain registered and keep working as the legacy secret-storage path, which the provider continues to fall back to (`getConfiguredApiKey` → secret storage, see `provideLanguageModelChatInformation`).

## Verification

- [x] Installed the packaged VSIX on VS Code 1.132 — `+ Add Models → OpenCode Go` now prompts and creates the group; gear/context-menu actions work; the leftover group is deletable.
- [x] `npm run compile` — clean.
- [x] `npm test` — 161/161 pass.
- [x] `npm run lint` — clean.

PR #125, opened 2026-08-10
fix(agents): OpenCode Go/Zen in the Agents window + remove providers from Language Models (#122)

### LINKED ISSUE #122: [BUG] OpenCode models missing from VS Code Agents window and "+ Add Models" menu
## 🐛 What happened?

OpenCode models (`OpenCode Go` / `OpenCode Zen`) work inside the GitHub Copilot Chat view, but do not appear in the VS Code Agents window.

## 🔄 Steps to reproduce

1. Open the VS Code Agents window.
2. Open the **Language Models** tab.
3. Click **+ Add Models**.
4. Observe that `OpenCode Go` and `OpenCode Zen` are missing from the provider list.

## ✅ What did you expect?

OpenCode models should be registered and selectable within the VS Code Agents window/session view.

## 🖥️ Environment

- **Extension version:** `0.5.1`
- **VS Code version:** `1.132.0`
- **Provider:** OpenCode Go / OpenCode Zen

Image

Fahad090NP:
Closes #122

## Problem

Two issues reported against the Agents window and Language Models management:

1. **Models missing from the Agents window.** VS Code ≥1.129 runs the Agents window in a separate agent host process and keeps the two mechanisms this feature depends on **off by default**:
   - `chat.agentHost.byokModels.enabled` (experimental) — the BYOK language-model bridge that mirrors extension BYOK models into agent-host sessions.
   - `extensions.supportAgentsWindow` — without it, code extensions are **disabled in the Agents window process entirely** (`canExecuteOnSessionsWindow` returns `false` for any extension with a `main` entry), so OpenCode Go/Zen never appeared in the Agents window model picker nor in its **+ Add Models** vendor list.
2. **No way to remove the providers.** `OpenCode Go` / `OpenCode Zen` always appeared in the Manage Language Models list and every model picker with no way to delete them (follow-up feedback on #122).

## Changes

**`fix(agents)` — Agents window support (commit `c5fbceb`):**
- On activation (when `opencodego.agentsWindow` and the new `opencodego.autoEnableAgentsWindow` setting are on), the extension auto-enables both core settings, merging with any existing user values:
  - `chat.agentHost.byokModels.enabled` → `true` (VS Code 1.129+).
  - `extensions.supportAgentsWindow."ltmoerdani.opencode-copilot-chat"` → `true`.
- Records in globalState exactly what it flipped and **reverts only its own changes** when the user disables `agentsWindow` (or turns off auto-config).
- Shows a one-time notification with a **Reload Now** button (both settings require a window/agent-host restart).
- Legacy agent-host providers (`targetChatSessionType: "copilotcli"`) remain registered for VS Code 1.125–1.128 where the bridge does not exist.

**`feat(models)` — remove providers from Language Models (commit `d4ee65d`):**
- New `opencodego.enabled` / `opencodezen.enabled` settings (default `true`); the vendor contributions carry matching `when` clauses, so a disabled provider disappears from the Manage Language Models view, the `+ Add Models` list, the Chat picker, and the Agents window (provider registration is skipped at startup).
- New commands: `OpenCode Go: Remove/Re-add Provider in Language Models` (same for Zen).
- New **Remove from Language Models** / **Re-add to Language Models** action in the Manage Provider QuickPick (reachable even without an API key).
- API keys and BYOK groups are kept, so re-enabling restores the provider unchanged. A window reload is required after toggling.

**`chore(tooling)` — linters/formatters honor `.gitignore` at runtime (commit `741cf45`):** ESLint reads `.gitignore` via the new shared `scripts/gitignore-patterns.mjs`, Prettier runs with `--ignore-path .gitignore`, markdownlint-cli2 config (renamed `.markdownlint-cli2.jsonc`) uses `"gitignore": true`.

**`chore` — ignore `tmp/` (commit `317eaf8`):** for temporary working copies; also excluded from the VSIX via `.vscodeignore`.

## Verification

- [x] Packaged VSIX tested on VS Code 1.132: after reload, OpenCode Go/Zen appear in the Agents window model picker and the **+ Add Models** list; both providers work in Copilot Chat.
- [x] Removing a provider via the command/QuickPick + reload removes it from Language Models, Chat picker, `+ Add Models`, and the Agents window; re-adding restores it with key/group intact.
- [x] `npm test` — 161/161 pass. `npm run lint` — clean.
- [x] VSIX contents verified (no `tmp/` packaged).


ltmoerdani:
Hi @Fahad090NP, nice work. I ran through the diff and the Agents window wiring is solid: the auto-enable + revert flow with globalState tracking is a clean way to handle settings we don't own, and the `when` clause on the vendor contributions is the right mechanism (docs confirm it controls the Manage Models list).

One thing I want to flag before we merge. In `activate()` and `warmModelPickerMetadata()` you read the Zen flag like this:

```ts
const opencodeCfg = vscode.workspace.getConfiguration("opencodego");
const zenProviderEnabled = opencodeCfg.get<boolean>("opencodezen.enabled", true);
```

`getConfiguration("opencodego")` resolves keys relative to that section, so this actually reads `opencodego.opencodezen.enabled`, which never exists. The setting is defined as `opencodezen.enabled`, so the read always falls back to `true`. Net effect: disabling Zen through the new command sets `opencodezen.enabled = false`, the `when` clause hides the vendor from Manage Models, but the provider is still registered at runtime, so Zen models stay in the Chat picker. That's the opposite of the "disappears from the Chat picker" claim, but only for Zen. Go is unaffected.

Fix is to read from the right section:

```ts
const zenProviderEnabled = vscode.workspace.getConfiguration("opencodezen").get<boolean>("enabled", true);
```

Same pattern in `warmModelPickerMetadata()` (~line 1089).

Minor: in `manage()`, agent vendors use `getConfiguration(this.definition.vendor)` with `opencodego-agent` / `opencodezen-agent`, so `.get("enabled")` there also always returns `true`. Not reachable through a real command right now, but worth guarding so it doesn't bite later.

Heads up on the merge: the branch is currently conflicting with `main` on `CHANGELOG.md` and `package.json` (main dropped `managementCommand` in #124, which is why the conflict shows up there). `src/extension.ts` merges clean. Resolution for the package.json conflict is just keeping the `when` clause you added. Happy to resolve it on your branch, or you can rebase yourself, whichever you prefer.

One design question, not a blocker: `autoEnableAgentsWindow` defaults to `true`, meaning the extension writes VS Code core settings (`chat.agentHost.byokModels.enabled`, `extensions.supportAgentsWindow`) on first activation. It's documented and revertible, but I'm curious if you think it should be opt-in instead.

Also, if you can add a quick test for the `enabled` gate, that'd be nice. It's config scoping, so it needs a small helper that can be mocked, but it's exactly the kind of bug that slips through otherwise.

Once the Zen fix lands I'm happy to merge. Thanks again for picking this up.

Fahad090NP:
Addressed all review points:

1. **Zen config scoping bug — fixed.**  in a new pure module (src/providerEnablement.ts) maps every vendor to the full root-configuration key; ,  and  now read from the root configuration, so Zen disables correctly (when-clause + registration + picker all consistent).
2. **Agent-vendor guard in manage() — fixed.** Agent variants resolve to their base vendor key.
3. **Test added.** src/test/providerEnablement.test.ts covers all four vendors incl. agent variants (10 new tests, 171/171 green).
4. **Rebased onto current main** (no more CHANGELOG/package.json conflicts; kept the `when` clauses).
5. **On the autoEnableAgentsWindow default:** I kept it `true` deliberately — `agentsWindow` already defaults to `true` and promises Agents-window support, so a silent no-op would be the bug. The writes are documented in the setting description, tracked in globalState, revertible on disable, and the user gets a one-time notification with a reload button. Happy to flip it to opt-in if you'd rather not touch core settings by default.

ltmoerdani:
Thanks for the thorough follow-up. `providerEnablement.ts` is a much cleaner home for this than inline config reads, and the root-config read is exactly what the bug needed. The regression test with the comment explaining section-scoped fallback covers the subtle part well.

I pulled the branch and ran it locally: compile is clean and `npm test` passes 171/171 including the new suite, so the fix checks out.

On `autoEnableAgentsWindow`: agreed, keeping it `true` is the right call. `agentsWindow` already defaults to true and promises Agents window support, so a silent no-op would be the actual bug here. The writes being documented, tracked in globalState, revertible on disable, and surfaced with a reload notification covers the downside well enough. No change needed.

Merge state is clean now, so I'll get this merged. Thanks again for the fast turnaround.

PR #123, opened 2026-08-10
fix(history): echo reasoning_content on multi-turn thinking requests

Fahad090NP:
## 📝 What does this change?

Fixes repeated `HTTP 400: The reasoning_content in the thinking mode must be passed back to the API` on multi-turn conversations with DeepSeek V4 (and other OpenAI-compatible reasoning models).

`convertMessage()` was silently dropping `LanguageModelThinkingPart` from assistant history, so follow-up turns omitted the previously emitted `reasoning_content` that DeepSeek requires to be echoed back unchanged. This fix extracts the thinking text from history and attaches it as `reasoning_content` on assistant messages.

## 🧪 How did you test it?

- `npm run compile` passes
- `npm run lint` passes (eslint + prettier)
- `npm

PR #126, opened 2026-08-11
test(reasoning): typeof guard + unit tests for reasoning history helpers (review notes on #123)

Fahad090NP:
Addresses the two non-blocker review notes on merged PR #123.

## Changes

**1. `typeof` guard for the proposed `LanguageModelThinkingPart` API**
`thinkingPartText()` used `part instanceof vscode.LanguageModelThinkingPart` without the `typeof ... === 'function'` guard that `streaming.ts` already uses for the same proposed API. The guard is now in place so the extension degrades gracefully on any hypothetical host without the API.

**2. Unit tests for the pure reasoning-history helpers**
`shouldEchoThinkingHistory()` and `thinkingPartText()` lived in `extension.ts` and had no tests. Extracted them into a new pure module `src/reasoningHistory.ts` (the value-normalization logic became `thinkingTextFromValue()`, keeping the `instanceof` check in the caller) and added a unit-test suite covering:

- `thinkingTextFromValue()` — plain strings, chunk arrays, non-string chunks, empty inputs.
- `shouldEchoThinkingHistory()` — every model family: DeepSeek/Kimi/GLM/Qwen/MiniMax (echo), Gemini (echo), MiMo/GPT/Claude (no echo, preserving the issue #38 carve-out), unknown models, and `undefined` model ids.

`src/extension.ts` behavior is unchanged — this is a pure extraction + defensive guard.

## Note

This branch also carries the `tmp/` ignore + `.gitignore`-aware tooling commits (identical content to PR #125). They're cherry-picked here only so lint passes on `main` today (which doesn't ignore `tmp/` yet); once either PR merges the other will reconcile cleanly. Happy to drop them and have the maintainer merge #125 first if preferred.

## Verification

- [x] `npm run compile` — clean.
- [x] `npm test` — 177/177 pass (was 161/161; +16 new).
- [x] `npm run lint` — clean (markdownlint now ignores `tmp/` via `gitignore: true`).


Fahad090NP:
**Proposal: keep `tmp/` in `.gitignore` as the designated scratch space for reference codebases**

While working on the Agents window models (#122) I needed the actual VS Code source to verify session-window/session-type behavior, so I dropped a full copy of `microsoft/vscode` into a local `tmp/` folder. That copy must never be committed or shipped:

- **`.gitignore`** — `tmp/` is ignored, so it can hold large reference trees (e.g. OpenCode's or VS Code's codebase) without polluting the repo or git status.
- **`.vscodeignore`** — `tmp/**` is excluded from the VSIX (a raw reference tree would otherwise balloon the package by hundreds of MB — the earlier packaged VSIX hit ~235 MB because of this).
- **Linters/formatters** — ESLint (via `scripts/gitignore-patterns.mjs`), Prettier (`--ignore-path .gitignore`) and markdownlint (`gitignore: true`) all derive their ignores from `.gitignore` at runtime, so `tmp/` is silently excluded everywhere.

Keeping reference copies of upstream codebases around is genuinely useful for verifying structural/architectural changes (e.g. how VS Code's `filterModelsForSession` / BYOK bridge behave across versions), and `tmp/` gives us a single, predictable place for it that all tooling already respects. I'd propose we keep this convention going forward.

Fahad090NP:
Hi @ltmoerdani, quick one on the two bot PRs sitting open for ~10 days:

- **#90** `chore(deps-dev): bump typescript from 6.0.3 to 7.0.2`
- **#91** `chore(deps-dev): bump @types/node from 26.1.0 to 26.1.2`

They're not doing anything while parked open — could we either accept or reject them? My take: accepting both is a good choice. They're dev-only dependency bumps (TypeScript 7 is already the toolchain we compile with locally — `tsc -p ./` passes, and the compiled output is identical), CI runs build + tests + GitGuardian on every PR, and we have the unit suite (177 tests) and linters as the safety net — plus my upcoming PRs follow strict linting anyway, so any regression would be caught before it ships.

Fahad090NP:
Correction to my previous comment: I said TypeScript 7 is already what we compile with locally — that's wrong. The local toolchain is TypeScript **6.0.3** (declared `^6.0.3`), so #90 is a **major** bump (6 → 7), not a no-op. My point stands though: CI (build + tests + lint + GitGuardian) runs on those PRs themselves, so accepting them is still safe — the PR's own checks would fail loudly if anything broke, and we have 177 unit tests + linters as the net. #91 (@types/node 26.1.0 → 26.1.2) is a routine patch bump.

Fahad090NP:
Hi @ltmoerdani, one more request: once these PRs (#125, #126 — and the two dependabot ones above) are approved/rejected, could we get a **new release published** with our changes?

Reasoning: **DeepSeek V4 is currently broken in the released version** (0.5.1) — the multi-turn `reasoning_content` 400 error (`The reasoning_content in the thinking mode must be passed back to the API`) that #123 fixed is still what everyone on the marketplace is hitting. DeepSeek V4 Flash is the most loved and widely used model in the OpenCode ecosystem, and the whole point of this extension is letting those users run it inside GitHub Copilot. Merging the fixes without releasing them means the bug keeps affecting users for another release cycle.

A quick cut after the pending PRs settle would unblock them.

ltmoerdani:
hey @Fahad090NP, just went through #126. the `typeof` guard matches what `streaming.ts` already does, the `reasoningHistory.ts` extraction is clean, and the +16 tests cover every model family including the issue #38 carve-out. merging today.

on the dependabot PRs: #91 (@types/node patch) is a quick yes, I'll approve it. #90 (TypeScript 6 → 7) I'm holding. major toolchain bumps need a local build pass first, so I'll handle that separately after the release cut.

release 0.5.2: agreed. DeepSeek V4 Flash being broken in the marketplace right now is the exact bug #123 fixed three days ago, and every OpenCode user on the extension is hitting it. cutting 0.5.2 is long overdue.

plan:
1. merge #126 + #91
2. bump 0.5.1 → 0.5.2
3. CHANGELOG (I'll handle)
4. tag + publish

shooting for today if both merges land clean.

PR #129, opened 2026-08-11
chore(tooling): strict-but-sane lint stack + intelligent pre-commit gate

Fahad090NP:
Brings the strict-lint branch up to date with main (incl. the 0.5.2 release cut) and finalizes the tooling:

## What's new in this PR

**1. `chore(eslint)`: keep type-aware strictness, drop style-only rules** (`ffce99d`)
- typescript-eslint `strict` + `strictTypeChecked` stay (the rules that catch real bugs: no-unsafe-*, no-unnecessary-condition, …).
- The pure-`stylistic` layer is removed — it fought prettier and produced churn without catching bugs (array-type, consistent-type-definitions, prefer-for-of).
- `restrict-template-expressions` allows numbers/booleans; `no-floating-promises` is off for `*.test.*` (node:test's runner handles those promises).
- Kept: yml/jsonc config linters, zero-tolerance `no-warning-comments`, `.gitignore`-derived ignores.

**2. `chore(lint)`: unit tests run as part of lint** (`a84402a`)
`npm run lint` / `bun lint` now ends with a Tests step (compile + unit tests): `✔ Editorconfig ✔ ESLint ✔ Markdown ✔ Prettier ✔ Shell ✔ TypeScript ✔ Tests`.

**3. `feat(hooks)`: intelligent pre-commit lint** (`a4ef449`, `6cf7714`)
New `scripts/staged-lint.js` (also `npm run lint:staged`). Instead of linting the whole tree on every commit (slow) or only the staged files (blind), it lints exactly what a change can affect:
- **ESLint**: staged JS/TS files **plus their direct dependents** (files that import a staged file, resolved from the real import graph) — so changing a module can never leave type-aware errors in its consumers.
- **markdownlint / editorconfig / shellcheck**: staged files.
- **tsc check + unit tests**: only when `src/`/`scripts/` changed.
- Formatting stays with lint-staged, running after the gate.

Measured: commit touching only config/hook files ≈ 1s; commit touching `src/` ≈ 10s. The full-tree lint stays in CI and on demand via `npm run lint`.

## Also included (branch history, previously unmerged)
- The strict tooling overhaul: unified `scripts/lint.js`/`format.js` runners (picocolors output), script renames (`*.mjs`/`*.mts` → `*.js`/`*.ts`), `editorconfig-checker` + `shellcheck` + `tsconfig.check.json` type-check, zero-tolerance formatting, dev-dependency bumps (eslint 10.8.1, @types/node 26.2), husky PATH fixes.
- Merge of current main (0.5.2 release cut incl. docs; conflicts resolved, merged docs re-formatted to pass the strict checks).

## Verification
- [x] `npm run lint` — all 7 steps green (incl. 180 unit tests).
- [x] Pre-commit gate exercised end-to-end (config-only commit ≈1s, src commit ≈10s; dependents detection verified).
- [x] Branch rebased/merged onto latest main — no conflicts.


Fahad090NP:
Hi @ltmoerdani — heads up before you dive in: **this is a huge PR** (96 files), so here's the recommended review path:

1. **Start in the File Changes section on GitHub** — the commit-by-commit view makes it digestible: the strict tooling overhaul is 8 commits, and the newer tuning (ESLint scoping, tests-in-lint, intelligent staged lint) is 4 commits on top. The diff of the *new* commits is small and self-contained; most of the 96 files are the repo-wide format pass from the earlier branch history.
2. **Then test locally**: `npm ci && npm run lint` (runs all 7 steps incl. tests), then make a small commit to feel the new pre-commit gate — a docs-only change commits in ~1s, a src change in ~10s, and the hook lints the staged files **plus the files that import them**.

A note on the history: I've been working on this branch for days and it is **fully up to date with the current codebase** (merged through the 0.5.2 release cut — no conflicts, verified mergeable). I deliberately started with the *strictest* possible stack (strict + strictTypeChecked + stylistic + yml + jsonc, everything an error), ran it over the whole tree, and then — after fixing what it surfaced — **dialed it back to what's actually worth keeping**: type-aware correctness rules stay, pure-stylistic rules are gone, test-file ceremony is gone. The goal is zero useless errors for future contributors, not maximal strictness.

Fahad090NP:
**What this PR fixes that we weren't even aware was broken**

Running the strict stack over the whole tree (and merging the newest main into it) surfaced a class of problems the previous tooling could not see:

1. **Type-unsafe plain-JS scripts.** `scripts/lint.js` and `format.js` only passed the old lint because nothing type-checked them — the `no-unsafe-*` rules had nothing to work with. The new `scripts/staged-lint.js` had to be fully JSDoc-typed before the gate accepted it; the existing scripts only got away with it by luck.
2. **Unnecessary conditionals and assertions that were hiding real sloppiness.** `vscode.version ?? ""` on a property the API types guarantee non-optional; `!visionModels ||` guards around a call that can never return null; an `as { value: unknown }` cast that changed nothing; a computed-key `delete` in the agents-window revert path (the kind of pattern that silently breaks under refactors).
3. **Formatting drift in shipped docs.** The 0.5.2 release-cut docs (5 files) were merged with prettier violations, and two older docs carried editorconfig trailing-whitespace violations — main's CI never checks those, so they would have shipped as-is forever. The merge here caught and fixed them.
4. **The consumer-blindness problem.** The old hook linted either everything (slow) or only the staged file (blind). A one-line change to a module could leave type-aware errors in every file importing it — undetectable until CI ran, or worse, until a contributor hit them locally. The staged lint resolves the import graph and lints dependents, so the guarantee now holds *at commit time*, not just in CI.
5. **Test-file ceremony.** `void it(...)`/promise noise everywhere told us the strictness was misfiring on the test runner — now scoped away instead of being worked around.

All of these are fixed by this branch; CI's old full-tree checks never would have caught 1, 3 or 4.

Fahad090NP:
**Review map — what deserves attention vs. what to skim**

If you want to keep the review focused, here's the priority order:

- `scripts/staged-lint.js` — the heart of the PR. Read it with the question "is this the right trade?" in mind: linting staged files + one level of dependents keeps commits fast while covering the real risk. The import-graph resolver is deliberately conservative (direct dependents only, relative specifiers).
- `eslint.config.mjs` — the philosophy. Only two rule relaxations (`restrict-template-expressions` with number/boolean allowance, `no-floating-promises` off for tests) plus dropping the `stylistic` layer. Everything else in the strict stack is untouched.
- `.husky/pre-commit` — now two lines of actual logic (staged lint gate, then lint-staged). The nvm PATH shim from the branch history stays, since it prevents silently skipped hooks.
- `.editorconfig` / `tsconfig.check.json` — the supporting casts: editorconfig is the cheap always-on guard; the check tsconfig type-checks `scripts/` which the build tsconfig never did.
- Everything else (script renames, deps bumps, doc formatting) is mechanical — safe to skim.

One honest caveat: `tsc -p tsconfig.check.json` and the unit tests run in the hook only when `src/`/`scripts/` changed. The full gate still runs in CI on every push, so nothing regresses silently — but if you'd rather have tests on every commit regardless, that's a one-line change in `staged-lint.js`.

Fahad090NP:
**Production problems these linting fixes prevent**

Some of the changes here are style — but several guard against failures that would only have shown up in production, after release:

1. **Non-null assertions removed from the extension context (`_extensionContext!` → guarded `extensionContext()`).** The old code dereferenced `_extensionContext!` in ~30 places — a compile-time lie. The type-checked lint bans that pattern, so it surfaced and got fixed properly: a call path that ever runs before activation (a Command Palette invocation during a slow activation, a settings-change handler firing mid-reload, a cached request arriving before model discovery) would have dereferenced `undefined` and crashed the extension host with an opaque `TypeError`. Now it throws a clear "extension context not initialized" error at the exact spot, fail-fast.

2. **`void`-marked fire-and-forget promises (`writeProfiles`, `writeActiveProfile`, …).** `no-floating-promises` forced explicit intent on every un-awaited promise. The real risk this removes: a silently dropped rejection — e.g. a profile-persistence write failing and nobody ever seeing the rejection, which reads in production as "my usage profile randomly reset".

3. **Unit tests now run inside `bun lint` / `npm run lint`.** Previously a "lint-clean" commit could carry broken tests — I literally hit two failing `reasoningHistory` tests during this work that the old lint would have waved through. One command is now the whole gate.

4. **Dependents are linted at commit time, not at CI.** The reviewer on #125 called the Zen config-scoping bug "exactly the kind of bug that slips through otherwise": a one-line change to a module leaves type-aware errors in every consumer. With the old hook those surfaced only after a push — and as the 0.5.2 release docs showed (merged with prettier violations main's CI never checks), the CI gap was real. The staged lint closes it at the moment of commit.

5. **Type-checked hook scripts.** `scripts/lint.js` / `format.js` / `staged-lint.js` run on every contributor machine. Untyped `spawnSync`/`readdirSync` calls could fail silently and make the hook *pass while linting nothing*. `tsconfig.check.json` now type-checks `scripts/` (the build tsconfig never did), and `staged-lint.js` is fully JSDoc-typed — the hook cannot silently degrade.

None of these would have been caught by the old full-tree lint — they're exactly the class of bug that ships, then shows up in a user's issue report weeks later.

ltmoerdani:
Reviewed locally. Pulled the branch, ran `npm ci && npm run lint`, all 7 steps green. Tested the staged gate too: docs-only commit ~0.6s, src commit ~8s, dependents detection works as described. Numbers check out.

The non-null assertion fix in `extension.ts` is the real win here. `_extensionContext!` in ~30 spots was a compile-time lie. Any code path hitting those before `activate()` finishes (slow activation, settings-change mid-reload, cached request before model discovery) would have dereferenced undefined and crashed the host with an opaque TypeError. The guarded `extensionContext()` accessor with fail-fast is the right call. Same goes for the `void`-marked fire-and-forget promises. Silent rejections on `writeProfiles` / `writeActiveProfile` reading as "usage profile randomly reset" in prod is exactly the kind of thing that's hell to debug after the fact.

Couple of things worth flagging, none blocking:

1. The "test-file ceremony is gone" framing in your comment doesn't quite match the diff. `no-floating-promises: off` for `*.test.*` makes the `void` prefix unnecessary (I checked, test files lint clean without it), but the branch *adds* 217 `void describe/it` and removes 0. So the ceremony is actually still there, just no longer required. Not a bug, just inconsistent with the narrative. Easy follow-up: drop the `void` prefix from test files in a separate pass, or leave them. Neither breaks anything.

2. Same vibe on "2 failing reasoningHistory tests fixed." The diff for `reasoningHistory.test.ts` is formatting + `void` additions only, no logic change to the assertions. If those tests were failing before this branch, the fix isn't visible in the diff. Maybe they got fixed by the strictTypeChecked rules surfacing something in `reasoningHistory.ts` itself? Just want to make sure I'm not missing context.

3. `no-warning-comments` bans `@ts-expect-error` repo-wide. Right now there's 0 usage so it's safe, but this extension leans on VS Code proposed API (`chatProvider.d.ts`, `languageModelThinkingPart.d.ts`) and those type gaps are real. When a future contributor needs to work around a proposed-API type mismatch, they'll have no clean escape hatch. Consider allowing `@ts-expect-error` specifically (keep `@ts-ignore` banned, that one's genuinely always wrong since it has no expiry).

The `staged-lint.js` design is solid. Conservative import-graph resolver (direct dependents only, relative specifiers), clear separation between "always run" (editorconfig) and "conditional" (tsc + tests on src/scripts change). The `tsconfig.check.json` addition closing the gap where `scripts/` was never type-checked is genuinely valuable.

Mergeable from my side once you've seen the notes above. Want me to merge with `--merge` (preserves all 15 commits), or you want to address the `void` cleanup first @Fahad090NP?

Fahad090NP:
Thanks @ltmoerdani — great catches, all three addressed. Two are now committed and pushed; one is a fair correction to my narrative.

**1. The `void` ceremony — agreed, and now actually gone.** You're right that the framing didn't match the diff: the branch *added* the prefixes and removed none. Since `no-floating-promises` is off for `*.test.*`, the prefix is pure ceremony — so instead of leaving it, I removed it. Commit `22e04b7` drops all 217 `void describe/it/test` prefixes across the 15 test files (mechanical pass, lint + tests still green). Now the diff actually matches the story.

**2. The "2 failing reasoningHistory tests" — fair, my comment was misleading.** Those two failures were a development-time artifact: my first draft of the new test file asserted `"think step 1"` instead of the actual `"think \nstep\n 1"` join behavior, and I fixed the expectations *before* the commit. From the PR's perspective the files are new and never existed in a failing state — there was no committed red that the lint rescued. The honest version of that point is only #3 in my earlier comment: `lint` didn't catch tests breaking, because `lint` never *ran* tests — which is exactly why it now does.

**3. `@ts-expect-error` — implemented.** Commit `5246434` removes it from the banned terms (keeping `todo`/`fixme`/`xxx`/`hack`/`@ts-ignore` banned). Your reasoning is exactly right for this codebase: the proposed-API type gaps are real, and unlike the never-expiring alternative, `@ts-expect-error` self-heals — the moment the proposed API ships, the directive fails and the error resurfaces instead of rotting.

Nothing else pending from my side — merge with `--merge` whenever you're ready, and thanks for the thorough review.

Fahad090NP:
Extension policy update on this branch: **latest standards with standard file names only** — `.ts`/`.js`, never `.mjs`/`.mts`, and TypeScript preferred over JavaScript everywhere. Implemented in two commits:

1. **`76570cc` — prefer TS over JS.** `eslint.config.ts` is now the ESLint config file (ESLint 10 loads TypeScript configs natively — `jiti` added as a devDependency). The four remaining JavaScript scripts became real TypeScript: `lint.ts`, `format.ts`, `staged-lint.ts`, `run-unit-tests.ts` (typed, not JSDoc-typed). `tsx` is a devDependency so the `.ts` scripts run on any Node (CI runs Node 20, which has no native type-stripping); `validate-models` and `test-retry` now use the local `tsx` instead of `npx --yes tsx`.

2. **`c817871` — consistency.** `.markdownlint-cli2.jsonc` → `.markdownlint-cli2.json` (no comments in the file, so `.json` is the honest extension); references updated across `package.json`, `lint.ts`, `staged-lint.ts` and `.vscodeignore`; the empty `.markdownlint/rules/` directory removed.

Net result for the review: the repo is now TypeScript (`.ts`) for all source and scripts, JSON for configs — zero `.js`, `.mjs`, `.mts` files anywhere. An earlier intermediate commit in this branch briefly used a CommonJS config; it was superseded by the TypeScript config and is irrelevant to the final state.

ltmoerdani:
Just pulled the latest state and re-ran the full check locally: `npm ci && npm run lint`, all 7 steps green on the 20-commit head. Also spot-checked the three points from your comment directly.

1. The `void` cleanup is confirmed gone. Grep for `void describe/it/test` across `src/test/` comes back 0, and lint + tests still pass without them. Diff now matches the story, nice.

2. The reasoningHistory clarification makes sense. I was looking at the merged-in diff and assuming there was a committed red; knowing it was a pre-commit expectation fix explains why the lint didn't "rescue" anything. Your point #3 is the honest framing, and fixing it (tests now run inside lint) is the actual value.

3. `@ts-expect-error` allowance is in, verified. `eslint.config.ts` line 106 keeps `@ts-ignore` banned while dropping `@ts-expect-error`, and there's a comment documenting the proposed-API rationale so future contributors don't re-litigate it.

Also checked the two newer commits (`.ts` for all scripts + `eslint.config.ts`, markdownlint config renamed to `.json`). Zero `.mjs`/`.mts` left in the repo, configs resolve cleanly, and the staged-lint still runs through `tsx` on Node 20 without a native type-stripping dependency. All good.

Mergeable from my side. I'll merge with `--merge` to preserve the full 20-commit history.

PR #132, opened 2026-08-12
feat(usage): sync Go usage from the official /zen/go/v1/usage endpoint (#130)

### LINKED ISSUE #130: [FEATURE] Sync Go usage from official /zen/go/v1/usage endpoint
## ✨ The feature

Have the Go usage status bar read Session (rolling 5h), Weekly, and Monthly meters from OpenCode's official `/zen/go/v1/usage` endpoint instead of estimating them locally. This makes the numbers account-wide and server-accurate.

## 🎯 The problem

Issue #23 reported drift from day one: the tracker showed "27d left" where opencode.ai said "19d left", and weekly/monthly percentages ran several points low. Root cause: OpenCode had no public usage API, so the extension had to estimate cost from tokens it saw and multiply by per-model pricing. That misses usage from the CLI/TUI, other devices, and anything before the extension was installed.

We shipped partial fixes: manual usage targets (#50) and a SQLite reader that pulls real billed amounts from the local `opencode.db` (#59/#60). They help, but SQLite is still per-machine, so cross-device usage stays out of the numbers.

Upstream just merged an official endpoint (anomalyco/opencode#16513): `GET /zen/go/v1/usage`. It authenticates with the same `opencode-go` key we already store in SecretStorage and returns rolling/weekly/monthly percent plus reset time computed server-side.

## 💡 Proposed solution

Add a sync layer in `goUsageTracker`:

1. On refresh, call `GET https://opencode.ai/zen/go/v1/usage` with the existing `opencodego.apiKey` from SecretStorage (same key and base URL as the models fetch).
2. Map the response onto `UsageSummary`:
   - `usage.rolling` → `session` (5h window, $12)
   - `usage.weekly` → `weekly` ($30)
   - `usage.monthly` → `monthly` ($60)
   - Use `resetsAt` for the "resets in" timer, which fixes the monthly anchor bug for real.
3. Keep Today / Yesterday and per-session spend from local tracking + SQLite, since the API does not return those.
4. Fall back to the current SQLite → tracked path when the API returns 401/403/404 or fails (offline, not live yet, no subscription).
5. Never persist or log the key, and keep the existing "device only" wording for the local numbers.

## 🔄 Alternatives considered

- Manual sync from opencode.ai, already shipped as "Set Usage Targets". Works, but needs a re-sync whenever usage happens elsewhere.
- OAuth browser flow to read the dashboard directly. The upstream PR makes this unnecessary.
- Rate-limit headers on ordinary responses. Not shipped yet, and a dedicated endpoint is cleaner.

## 📋 Additional context

- Upstream PR: anomalyco/opencode#16513 (merged to `dev` Aug 11 2026; confirm it is live in production before building on it).
- Follows the same split as robinebers/openusage#1095: API for account-wide limits, local history for spend.
- Related: #23, #50, #59, #60.
- The extension already talks to `https://opencode.ai/zen/go/v1/*` for models and chat completions, so no new auth surface is needed.

Fahad090NP:
Closes #130

## What

The status-bar Session/Weekly/Monthly meters were estimated locally (tokens × pricing + local `opencode.db`), which drifted from opencode.ai because CLI, cross-device and pre-install usage was invisible (issue #23). This syncs the meters from the official endpoint instead.

**Sync layer (`src/goUsageSync.ts`, pure + unit-tested):**
- `fetchGoUsage(apiKey)` — `GET /zen/go/v1/usage` with the existing Go key as a Bearer token; failures classified (unauthorized / no-subscription / not-found / network / invalid) so callers fall back to local estimates. The key is never logged or persisted.
- `mergeServerUsage()` — overlays server rolling/weekly/monthly percent + `resetsAt` onto the local summary; `spent` is derived from the authoritative percent. Today/Yesterday/per-session spend stay device-local.

**Tracker wiring (`src/goUsageTracker.ts`):**
- `syncServerUsage(apiKey)` with a 60s TTL (failures paced too, so a bad key can't hammer the API); `getSummary()` stays synchronous and overlays the cached snapshot.

**Extension wiring:**
- Startup sync with the stored key; per-request re-sync using the exact key the request ran under (covers native BYOK group keys); background sync on status-bar refresh with repaint on new data.

**Dialog fixes found along the way:**
- "Reset tracked usage data" was never wired — picking it did nothing. Now: modal confirm → `tracker.clear()` (also clears session costs) → repaint.
- Reset previously collapsed the card into the first-run "No data" state; a persisted `everTracked` flag keeps the card with zeroed local values instead.
- Quick-pick "Open OpenCode console" was also dead — now opens the console; a "Synced from opencode.ai" note appears when server meters are active.
- Usage panel + hover card: stable geometry (fixed width/columns, no layout jumps), consistent 14/16px gutters, always-visible Today/Yesterday rows; the webview is a clean compact card.
- Resolved the always-true/false TS hints (redundant cancellation guards, dead `if (tracker)`, and two wrong types: `GO_MODEL_PRICING` and the metadata `providers` map are genuinely partial at runtime).

## Verification

- [x] Endpoint verified live in production (401 shape matches upstream route source; upstream anomalyco/opencode#16513 merged).
- [x] `npm run lint` — all 7 steps green; `npm test` 188/188.
- [x] Branch merged with current main (incl. the strict-lint toolchain) — mergeable, no conflicts.


ltmoerdani:
Nice one @Fahad090NP, this is a clean implementation. I ran the branch locally before commenting: compile, the strict tsc check, and the full unit suite all pass (188/188), and the 8 new goUsageSync tests hold up. The endpoint was verified live, the fallback-to-local path is the right call, and wiring the two dead quick-pick actions (reset + open console) is a nice catch.

Two things I want to poke at before merge.

1. Multi-profile sync key. `refreshGoUsageStatusBar()` and the startup sync in `activate()` both read `secrets.get(SECRET_KEY)` and push that key onto the active tracker. With a single account that's the same key, no problem. With multiple profiles, switching profiles without a request in between pulls the wrong account's meters onto the active profile. And when a request does land, the status-bar sync and the per-request sync race, both write the snapshot, so the wrong-key result can win. The request path already resolves the exact key, so maybe the refresh path can do the same instead of reading `SECRET_KEY`. Am I missing something in how profile keys resolve here?

2. `hasData` ignores the server snapshot. The status bar returns "OpenCode Go" and the tooltip says "No usage data yet" whenever there's no local tracking, even when the server just reported real account-wide meters. That's basically the exact case #130 is about: someone with CLI usage installs the extension fresh and sees nothing until they send their first request. The quick pick and webview show the meters fine, so it's only the status bar and tooltip. Flipping `hasData` when a server snapshot exists would fix it.

Minor: `syncServerUsage()` has no in-flight dedup, so startup and a status-bar refresh at the same moment can fetch twice. Benign, last write wins, but a cheap guard would tidy it up.

No blockers from me. Happy to merge once you've had a look, and I'll keep it as a merge commit so your history stays intact.

Fahad090NP:
Let me do some more changes. I opened this PR too early. Sorry for that.

Fahad090NP:
All three points addressed and pushed (commits `172c2e9`, `98efb6b`, `0bff8b7`):

1. **Multi-profile sync key — fixed.** The extension now remembers which API key owns each profile fingerprint (`ensureProfileForApiKey` records it), and the status-bar background sync uses the **active profile's own key**, falling back to the extension secret only when unknown. Switching profiles without a request now syncs the right account's meters, and the refresh path no longer races the per-request sync with a different key (both resolve the same key for the same tracker).

2. **`hasData` ignoring the server snapshot — fixed.** `mergeServerUsage` now flips `hasData` to `true` whenever a server snapshot is applied, so a fresh install with CLI usage shows the real meters on the status bar and tooltip instead of "OpenCode Go" / "No usage data yet". Regression test added (189/189 passing).

3. **In-flight dedup — fixed.** Concurrent `syncServerUsage` calls for the same key share a single in-flight promise (startup + status-bar refresh no longer double-fetch); failures stay TTL-paced as before.

Ready to merge whenever you are — merge commit works for me.

ltmoerdani:
Nice one, all three fixes check out. I pulled the branch and ran it before commenting: compile, strict tsc, and the full suite all pass at 189/189. The new fresh-install regression test covers exactly the case I was worried about.

The profile-key map is a clean solution, and sharing one in-flight promise for concurrent syncs reads well. Flipping `hasData` in `mergeServerUsage` is minimal and does the job.

One heads-up rather than a blocker: the PR picked up two files outside the usage scope (`.husky/pre-commit` and the lint-staged config in `package.json`, now check-only). It makes sense as its own follow-up, and it's consistent with how we treat the pre-commit gate, so I'm fine keeping it in this PR. Just flagging so it stays intentional.

Merge commit as usual.

PR #133, opened 2026-08-12
fix(streaming): echo reasoning_content for thinking-off requests (DeepSeek 400)

### LINKED ISSUE #134: [Bug] DeepSeek V4 multi-turn 400 'reasoning_content must be passed back' when thinking is OFF
## 🐛 What happened

With a DeepSeek V4 model and **thinking turned off**, follow-up turns fail with a 400:

> Sorry, your request failed. Please try again.
>
> Client Request Id: fee2775e-0c6a-470c-9d8a-a10fec30aafe
>
> Reason: OpenCode Go API request failed (400) model=deepseek-v4-flash payloadBytes=419301: Error from provider (Console Go): Upstream request failed: [invalid_request_error] The reasoning_content in the thinking mode must be passed back to the API.
>
> at streamOpenCodeResponse (…/streaming.js:373) · at streamChatCompletions (…/streaming.js:64) · at OpenCodeProvider.provideLanguageModelChatResponse (…/extension.js:1993)

The #123 fix covers turns where the model's reasoning is surfaced as a thinking part. When thinking is **OFF**, the Go gateway wraps the response in `reasoning_content` (gateway bug #37635) and the extension emits it as **visible text** — never as a thinking part — so the history has nothing to echo back and DeepSeek's validator 400s on the next turn.

## 🔄 Steps to reproduce

1. Extension 0.5.2+, model `deepseek-v4-flash` (OpenCode Go), thinking set to **off**.
2. Send a first message (works, response may appear as text only).
3. Send a follow-up message in the same conversation.
4. Follow-up fails with the 400 above.

## ✅ What did you expect

Follow-up turns succeed with thinking off, just like with thinking on (the previous reasoning is echoed back as `reasoning_content`).

## 🔧 Fix

PR #133 — when thinking is off, the reasoning that was surfaced as visible text is now carried into the conversation as an internal marker data part (`application/vnd.opencode.reasoning+json`), and the next turn echoes it back as `reasoning_content`.

## 📎 References

- #123 — original reasoning_content echo fix (thinking-on path)
- PR #133 — this fix

Fahad090NP:
Fixes the recurring `400 The reasoning_content in the thinking mode must be passed back to the API` that fires **when thinking is OFF** for DeepSeek V4 (the #123 fix only covered turns where the reasoning was surfaced as a thinking part).

## Root cause

With thinking OFF on the Go gateway (`/zen/go/`), the extension applies the gateway workaround for bug #37635: the gateway wraps responses in `reasoning_content`, so the text is emitted as **visible `LanguageModelTextPart`** — never as a `LanguageModelThinkingPart`. The conversation history therefore contains **no thinking part** to echo, and the next turn omits `reasoning_content` on the assistant history message → DeepSeek's validator 400s.

## Fix (4 commits)

1. **`aa124db`** — new internal data part `application/vnd.opencode.reasoning+json` (`src/chatParts.ts`): a transcript-resident marker carrying the reasoning text.
2. **`3d7f277`** — the OpenAI extractor accumulates the reasoning it emitted as visible text and, after the stream completes, reports the marker part into the transcript.
3. **`7a98bc4`** — `convertMessage` reads the marker on assistant history messages and includes it in the echoed `reasoning_content` (alongside genuine thinking parts), closing the loop for the next turn.
4. **`2d70849`** — unit tests for the marker helpers (round-trip, MIME discrimination, malformed payloads) — 183/183 passing.

The marker is scoped to the `treatReasoningAsContent` path (Go gateway + thinking off), so nothing changes for thinking-on turns (genuine thinking parts) or other transports.

## Verification

- [x] `npm test` 183/183, `npm run lint` (strict, incl. tsc check) green.
- [x] Scoped: no behavior change when thinking is on or for non-Go transports.

Closes #134

### DIFF
--- a/src/chatParts.ts
+++ b/src/chatParts.ts
@@ -3,6 +3,7 @@ import { hasUsageSnapshot, toProviderUsagePayload, type UsageSnapshot } from "./
 
 export const OPENCODE_USAGE_DATA_MIME = "application/vnd.opencode.usage+json";
 export const COPILOT_USAGE_DATA_MIME = "usage";
+export const OPENCODE_REASONING_DATA_MIME = "application/vnd.opencode.reasoning+json";
 
 export function createUsageDataPart(usage: UsageSnapshot): vscode.LanguageModelDataPart | undefined {
   return createUsageDataParts(usage)[0];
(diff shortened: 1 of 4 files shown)


ltmoerdani:
Nice one @Fahad090NP, this closes the gap properly. I pulled the branch into a worktree and ran it locally:

- `npm run compile` clean
- `npm test` 183/183 (the 3 new marker tests are a good add)
- CI green on both checks

I traced the scoping and it holds up. `flushReasoningMarker()` only fires inside `streamChatCompletions`, and it's gated on `treatReasoningAsContent` so it can't leak into the thinking-on path or the other transports. The MiMo carve-out from issue #38 is safe too: even if a MiMo model hits the marker on history, `shouldEchoThinkingHistory` returns false and the tool-call branch has its own `shouldOmitReasoningEcho` gate, so the reasoning never gets echoed back to Xiaomi's validator.

One small thing, not a blocker: `isInternalDataPart()` in chatParts.ts still only covers the usage MIMEs. The new `OPENCODE_REASONING_DATA_MIME` isn't in there, so the marker slips past the `isInternalDataPart` check in `convertMessage` and also gets counted toward the token estimate in `estimateDataPartTokenCount` / `collectRequestParts`. Functionally it's fine because the marker branch in `convertMessage` runs first and hits `continue` before that check, and the reasoning text was already emitted as a text part anyway so the token count isn't really off. Just a consistency thing, would be nice to add the MIME to `isInternalDataPart` so the guard is uniform. Happy to merge without it though.

The one thing only a live run can confirm is the actual 400 disappearing on a 2-turn DeepSeek conversation with thinking off, like you noted in #134. The mechanism is sound from tracing the code, but I'd want to verify it once on a real Go-gateway session before cutting a release.

I'll merge this with a merge commit so your commits stay intact.

Issue #106, opened 2026-08-05
[BUG] Duplicate model in OpenCode Zen model list


CoderTCY:
## 🐛 What happened?

All models in OpenCode Zen shows twice in model list

## 🔄 Steps to reproduce

1. Install ext
2. Open model list

## ✅ What did you expect?

All models only show at once

## 🖥️ Environment

- **Extension version:** 0.5.0
- **VS Code version:** 1.131.0
- **Provider:** OpenCode Zen
- **Model:** all

## 📸 Logs or screenshots (optional)

<img width="320" height="332" alt="Image" src="https://github.com/user-attachments/assets/fc3e2952-7f24-486f-be3c-38154d4122b6" />

```
# OpenCode Zen Diagnostics

## Recent Requests

### 1. deepseek-v4-flash-free

- time: 2026-08-05T02:34:39.117Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 165846
- ttfbMs: 3399
- totalBytes: 4506568
- totalEvents: 16962
- tokens: prompt=43748, completion=17056, total=60804, cached=37760
- cacheHitRatio: 86.3%
- finishReason: stop
- requestId: req-8bee7e9a
- sessionId: vscode-c7de75c4
- url: https://opencode.ai/zen/v1/chat/completions

### 2. deepseek-v4-flash-free

- time: 2026-08-05T02:31:26.364Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 228608
- ttfbMs: 2504
- totalBytes: 6007230
- totalEvents: 22596
- tokens: prompt=20823, completion=22828, total=43651, cached=0
- cacheHitRatio: 0.0%
- finishReason: tool_calls
- requestId: req-94bc4f9a
- sessionId: vscode-c7de75c4
- url: https://opencode.ai/zen/v1/chat/completions

### 3. deepseek-v4-flash-free

- time: 2026-08-05T02:27:48.841Z
- endpoint: chat-completions
- initiator: github.copilot-chat
- metadataSource: models.dev
- status: 200
- durationMs: 11166
- ttfbMs: 1695
- totalBytes: 276982
- totalEvents: 1044
- tokens: prompt=342, completion=1041, total=1383, cached=0
- cacheHitRatio: 0.0%
- finishReason: stop
- requestId: req-ae8073fb
[...]

Issue #86, opened 2026-07-29
[BUG] Non-agent OpenCode Zen returns 0 models when API key set via extension command (not native BYOK Add Models)


Witchcraft2k:
## 🐛 What happened?

After setting the OpenCode API key through the extension's own command (`OpenCode Go: Set API Key`), the **non-agent `opencodezen` provider** shows **0 models** in `vscode.lm.selectChatModels({ vendor: "opencodezen" })`, so Zen free models never appear in the Chat model picker dropdown.

Meanwhile:
- `opencodego` (non-agent Go): ✅ 23 models visible
- `opencodego-agent`: ✅ 23 models visible
- `opencodezen-agent`: ✅ 7 models visible
- **`opencodezen` (non-agent Zen): ❌ 0 models visible**

## 🔄 Steps to reproduce

1. Install the extension
2. Set your OpenCode API key via the command **`OpenCode Go: Set API Key`** (NOT via the native VS Code **Language Models → Add Models** BYOK flow)
3. Run **`OpenCode: Model Picker Diagnostics`** — observe `vendor: opencodezen` shows `models: 0`
4. Run **`OpenCode Zen: Diagnostics`** — observe `Models visible through vscode.lm.selectChatModels({ vendor: "opencodezen" }): 0`

Expected: The non-agent Zen provider should also find the API key from secret storage and return models, just like the agent variant does.

## ✅ Root cause

In `src/extension.ts`, the `provideLanguageModelChatInformation` method has this API key fallback logic:

```ts
// Line ~1825 — the guard that blocks secrets fallback
if (!apiKey && (this.definition.isAgentVariant || options.configuration)) {
  apiKey = await this.context.secrets.get(SECRET_KEY);
}

if (!apiKey) {
  return [];  // <- Zen stops here
}
```

The condition `(this.definition.isAgentVariant || options.configuration)` means the secrets fallback is **only entered** when:
- The provider is an **agent variant** (`isAgentVariant=true`), OR
- `options.configuration` is **truthy** (VS Code passed a BYOK configuration object, even if empty `{}`)

When the user stores the API key via the extension command (`context.secrets.store`) rather than through VS Code's native **Language Models → Add Models** BYOK flow, VS Code has no BYOK data for the `opencodezen` provider. So it calls `provideLanguageModelChatInformation` with `options.configuration = undefined`.

For non-agent providers with `options.configuration = undefined`:
- `!apiKey` = `true`
[...]

PR #135, opened 2026-08-12
fix(provider): duplicate OpenCode models after per-model config (e.g. reasoningEffort) (#131)

### LINKED ISSUE #131: [Bug] Duplicate OpenCode Zen / Go models after changing per-model reasoning effort
## 🐛 What happened

After setting a per-model option (e.g. `reasoningEffort`) in the model picker, every
OpenCode Zen (and Go) model is listed twice. Removing the per-model config
entry returns the list to a single set; changing the option again brings the
duplicates straight back.

Reported originally in #106 by @xianhongtao. That issue was closed as fixed in
PR #108 (0.5.2), but this is a separate trigger: the 0.5.2 fix only covers
provider groups that carry an `apiKey`. A per-model configuration group carries
no key, so the duplicates persist on 0.5.2.

## 🔄 Steps to reproduce

1. Install extension 0.5.2, set the Zen (or Go) API key via the extension
   command (`OpenCode Zen: Set API Key`), so the key lives in SecretStorage.
2. Reload the window, open the model picker and count Zen models (baseline).
3. Change `reasoningEffort` on one Zen model in the picker submenu.
4. Reopen the picker and count again.

Reported numbers (VS Code 1.132.1, extension 0.5.2):

| Step | Count |
| ---- | ----- |
| Reload (no per-model config) | 7 |
| Delete per-model group via Manage Models (Cog) | 7 |
| Change reasoning effort (group recreated) | 14 |

7 unique Zen models, doubled to 14 once a per-model config group exists.

## ✅ What did you expect?

Each model listed exactly once regardless of per-model configuration.

## 🖥️ Environment

- Extension version: 0.5.2
- VS Code version: 1.132.1 (System setup, Stable)
- Provider: OpenCode Zen (also affects OpenCode Go)
- API key set via: extension command (`Set API Key`)

## 🧠 Root cause (from `chatLanguageModels.json`)

`chatLanguageModels.json` contains a per-model configuration group for each
vendor that only has `settings`, no `apiKey`:

```jsonc
{
  "name": "OpenCode Zen",
  "vendor": "opencodezen",
  "settings": {
    "opencodezen:deepseek-v4-flash-free::session-2026-05-21-b::sk-***": {
      "reasoningEffort": "max"
    }
  }
}
```

VS Code calls `provideLanguageModelChatInformation` once per configured group
for a provider that declares a `configuration` schema. `_resolveConfiguration`
on a settings-only group yields `{}` (no `apiKey`). In the extension:

- Groupless call → falls back to SecretStorage → serves N models.
- Per-model group call (`configuration: {}`) → `getConfiguredApiKey` is
  undefined → the #106 BYOK-group flag is never set → SecretStorage fallback
  runs again → serves N more models.

Result: N × 2 while a per-model config group exists. The #106 fix (#108) only
suppresses the groupless call when an `apiKey`-bearing group call has been
observed, so it does not cover this path.

## 🔧 Proposed fix

In `provideLanguageModelChatInformation`, treat a non-undefined `configuration`
without an `apiKey` as a per-model configuration group and return `[]`:

```ts
let apiKey = getConfiguredApiKey(opts);
if (apiKey) {
  await this.markByokGroupConfigured();
} else if (opts.configuration !== undefined) {
  return []; // per-model config group carries no key; groupless call already serves models
}
```

This is safe for all known scenarios: #86 (key via command, no group), #106/#63
(groups that carry an `apiKey`), and agent variants (no `configuration` schema,
so `configuration` is always `undefined`).

## 📎 References

- #106 — original duplicate-model issue (fixed for apiKey-bearing groups in #108)
- #86 — SecretStorage fallback that makes this duplication possible
- #128 — duplicate report closed as dup of #106 (same root cause)
- Reporter diagnostics: #106 comment (2026-08-12) with `chatLanguageModels.json` contents

Fahad090NP:
Closes #131

## Problem

Setting a per-model option (e.g. `reasoningEffort`) in the model picker creates a settings-only group in `chatLanguageModels.json`. VS Code then calls `provideLanguageModelChatInformation` once per configured group, and the settings-only group resolves to `configuration: {}` — no `apiKey`. The extension's SecretStorage fallback ran on **both** the groupless call and the group call, serving every model twice (7 → 14 in the report).

The #106 fix only suppresses the groupless call when an `apiKey`-bearing group call was observed, so it doesn't cover this path.

## Fix

A group call whose `configuration` is present but carries no API key is now treated as a per-model configuration group and returns `[]` — the groupless call already served the models. Verified against the VS Code source (`languageModels.ts`): the groupless call passes no `configuration`, group calls always do, so this can't suppress the groupless path.

Per-model settings (e.g. `reasoningEffort`) are applied at request time via `modelConfiguration` and are unaffected.

## Verification

- [x] `npm test` 180/180, `npm run lint` (strict + tsc) green.
- [x] Scenarios preserved: #86 (key via command, no group), #106/#63 (groups carrying an `apiKey`), agent variants (no configuration schema → `configuration` always `undefined`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,11 +32,13 @@ jobs:
       - name: 🔎 Lint source and maintenance docs
         env:
           # The `shellcheck` npm package downloads its binary from GitHub
-          # releases on first run, which flakes in CI (socket hang up / rate
-          # limits). Use the runner's system binary instead.
-          SHELLCHECKJS_BIN: /usr/bin/shellcheck
+          # releases on first run (flaky in CI), and its SHELLCHECKJS_BIN
+          # validator requires a user-writable binary. Install via apt and
+          # point it at a writable copy in the runner's temp dir.
+          SHELLCHECKJS_BIN: ${{ runner.temp }}/shellcheck
         run: |
           sudo apt-get update && sudo apt-get install -y shellcheck
+          cp /usr/bin/shellcheck "$SHELLCHECKJS_BIN"
           npm run lint
 
       - name: 🎨 Check formatting
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,7 +30,14 @@ jobs:
         run: npm run compile
 
       - name: 🔎 Lint source and maintenance docs
-        run: npm run lint
+        env:
+          # The `shellcheck` npm package downloads its binary from GitHub
+          # releases on first run, which flakes in CI (socket hang up / rate
+          # limits). Use the runner's system binary instead.
+          SHELLCHECKJS_BIN: /usr/bin/shellcheck
+        run: |
+          sudo apt-get update && sudo apt-get install -y shellcheck
+          npm run lint
 
       - name: 🎨 Check formatting
         run: npm run format:check
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -2049,6 +2049,15 @@ class OpenCodeProvider implements vscode.LanguageModelChatProvider<OpenCodeModel
     // (issue #106, see step 2 below).
     if (apiKey) {
       await this.markByokGroupConfigured();
+    } else if (opts.configuration !== undefined) {
+      // A group call with a non-undefined configuration that carries no API
+      // key is a per-model configuration group (only `settings`, no key —
+      // e.g. a `reasoningEffort` picked in the model picker). VS Code
+      // resolves its configuration to `{}` here. The groupless call already
+      // served the models via SecretStorage, so serving them again would
+      // duplicate every model (issue #131). The per-model settings still
+      // apply at request time via `modelConfiguration`.
+      return [];
     }
 
     // 2. Fall back to the extension's own secret storage when BYOK did not


Fahad090NP:
To be very honest, I was unable to recreate this issue, but I did the possible fixes that break nothing. At least as I think.

ltmoerdani:
Thanks for this @Fahad090NP . dug into the root cause and your fix is spot on.

the issue only shows up when the key is set via the extension command (SecretStorage) AND the user picks a `reasoningEffort` in the picker. that second action creates a settings-only group in `chatLanguageModels.json` with no `apiKey`, so VS Code calls `provideLanguageModelChatInformation` twice (once groupless, once for the group) and our SecretStorage fallback served the full model set both times. the #106 fix only kicked in when a call carried an actual key, so it missed this path entirely.

your `else if (opts.configuration !== undefined)` discriminator is the right call. the groupless pass always has `configuration: undefined`, every group pass carries a `configuration` object (even `{}` for settings-only), so there's no risk of suppressing the groupless path. confirmed against the VS Code source.

two things worth noting from my side:

1. per-model settings (`reasoningEffort`, `contextSize`) are applied at request time via `modelConfiguration` in `provideLanguageModelChatResponse`, not during discovery. so returning `[]` from the group call doesn't drop those settings. verified this on our end.

2. the CI shellcheck change is a solid bonus. the npm package has been flaky on and off, your apt workaround is cleaner.

no blockers from me. all scenarios check out: #86 (key via command, no group), #106/#63 (native BYOK groups with key), agent variants (no configuration schema, so `configuration` stays undefined). merging as a merge commit to keep your commits intact.

not being able to reproduce: totally fair, the setup is specific (key via extension command + per-model config pick). a few of us had to trace the VS Code source to confirm the call pattern. the fix holds up regardless.

PR #136, opened 2026-08-12
feat(autocomplete): inline code suggestions with thinking off (experimental) (#49)

### LINKED ISSUE #49: [FEATURE] Add inline code suggestions with selectable AI model (Copilot-like experience)
## ✨ The feature
Add inline code suggestions (ghost text / Copilot-style) similar to GitHub Copilot and other extensions like Continue. The idea is to provide real-time completions while the user types, directly inside the editor.

Also allow selecting which AI model powers these suggestions.

## 🎯 The problem
Right now the extension is mostly chat-based, which breaks the coding flow and is slower compared to inline suggestions used in tools like Copilot or Continue.

## 💡 Proposed solution
Implement inline completions that appear as ghost text while typing, similar to Copilot/Continue behavior.

Add a setting to choose the model used for suggestions, for example:
```json
{
  "inlineSuggestions": true,
  "inlineSuggestionsModel": "deepseek-v4-flash"
}

Fahad090NP:
Closes #49 — **inline code suggestions (experimental, opt-in).**

## What

Ghost-text inline completions (Copilot-style) powered by the OpenCode gateway, with **thinking forced off**.

**Engine decision (live-measured, not assumed):** OpenCode exposes no FIM endpoint (`/completions` probed 404 on `zen/go`/`zen`/`beta`), and every opencode-go model is reasoning-first. Live probe results:

| Model | TTFB | Total | Hidden reasoning |
|---|---|---|---|
| deepseek-v4-flash (no reasoning_effort) | 1813ms | 2264ms | 128 chars — non-viable |
| **qwen3.5-plus (`enable_thinking=false`)** | **1518ms** | **1608ms** | **0 — genuine non-thinking** |

So the default engine is **qwen3.5-plus with `enable_thinking: false`** — zero hidden reasoning, ~1.5s time-to-first-token.

## Implementation (`src/autocomplete/`, 15 unit tests, 207 total at merge base — now 276 with #138)

- **context.ts** — bounded prefix/suffix window (10 lines before cursor, short suffix).
- **prompt.ts** — fill-in-the-middle emulation with FIM tokens over `/chat/completions` (Qwen family sends `enable_thinking: false`).
- **throttle.ts** — 300ms debounce, latest-wins, superseded runs aborted; the window is re-read live so `inlineSuggestionsDebounceMs` changes apply without a reload.
- **engine.ts** — streaming SSE parser with 3s timeout and keystroke abort; failures are logged with the exact reason to the **OpenCode Completions** output channel (network error / non-OK status / stream interruption).
- **provider.ts** — `InlineCompletionItemProvider` registering on `**`, gated by the opt-in setting; API key resolved with the same profile-first fallback as the chat path.
- **index.ts** — registration + settings (debounce, timeout, max tokens, prefix lines, suffix chars — all config-driven).

## Settings (all optional)

- `opencodego.inlineSuggestions` — opt-in (default `false`).
- `opencodego.inlineSuggestionsModel` — default `qwen3.5-plus` (verified non-thinking only).
- `opencodego.inlineSuggestionsDebounceMs` / `TimeoutMs` / `MaxTokens` / `PrefixLines` / `SuffixChars`.

## Review notes (addressed)

- API-key resolution: profile-first fallback (multi-profile / BYOK users).
- Engine failures are logged, never silently swallowed.
- `cleanCompletion` preserves leading newlines so nested-block completions keep their indentation.
- Autocomplete spend is not yet wired into the usage tracker — the per-day **Suggested / Approved** counters ship in #138; cost attribution is a documented follow-up.
- `qwen3.7-plus` added to the bundled fallback lists in #138.

## Testing

`npm test` (207 at the merge base, 276 with #138 stacked), strict ESLint + tsc + prettier + markdownlint all green. VSIX packages and installs cleanly.


Fahad090NP:
👋 **Calling for testers** — this is a draft, help us validate it:

**To test:**
1. Set `"opencodego.inlineSuggestions": true` in settings, reload the window, ensure your API key is set.
2. Open a JS/TS file, type code — ghost text should appear within ~1.5–2s.
3. Report: does the ghost text feel usable? Does it abort cleanly when you keep typing? Any errors in the output channel?

**What we know:** qwen3.5-plus with thinking forced off is the engine (measured ~1.5s TTFB, zero hidden reasoning). DeepSeek is measurably slower (hidden reasoning even "off").

**Known limitation:** our provider runs alongside Copilot's ghost text — if Copilot is active it may win the slot. Feedback on that UX is welcome.

Fahad090NP:
**Running alongside Copilot's own suggestions**

If GitHub Copilot is active, its ghost text / next-edit suggestions can win the slot over ours. You can turn them off and let OpenCode's suggestions take over:

- Ghost text: `"editor.inlineSuggestions.enabled": false`
- Next Edit Suggestions: `"github.copilot.nextEditSuggestions.enabled": false`

Both are user settings — no extension change needed. If you prefer Copilot's completions, simply leave `opencodego.inlineSuggestions` off and nothing changes.

Fahad090NP:
**Calling all testers — every finding is welcome**

Please test and report **anything** that looks off: wrong completions, latency spikes, ghost text that doesn't appear or doesn't go away, errors in the output channel, weird interactions with Copilot, config edge cases — big or small. **Do not leave even a small typo in the docs or settings descriptions unreported — I'll fix it.**

When reporting, please include:
1. The language + the code snippet around the cursor.
2. Your `opencodego.inlineSuggestionsModel` value (default `qwen3.5-plus`).
3. Approx. time from typing to ghost text appearing.
4. Anything from the **OpenCode** output channel around the failure.

Fahad090NP:
**Design notes for reviewers/testers**

- **Engine choice was measured, not assumed** — the probe (`scripts/probe-completion-latency.ts`, key from env only) showed `qwen3.5-plus` with `enable_thinking=false` at ~1.5s TTFB with zero hidden reasoning, while `deepseek-v4-flash` burned 128 reasoning chars even with thinking off. DeepSeek is intentionally not the default.
- **Cancellation is strict**: 300ms debounce, latest keystroke aborts the in-flight request (AbortSignal), 3s timeout. If you see stale ghost text, that's a bug — report it.
- **Context window is tiny by design** (10 lines before cursor + short suffix) — longer context will be considered if testers feel completions lack context.
- **New settings are opt-in**: `opencodego.inlineSuggestions` (default `false`) + `opencodego.inlineSuggestionsModel`.
- The `suffix` field is tolerated by the gateway (probed HTTP 200) but not relied on — FIM is emulated with tokens inline, so it works even if the gateway ignores the field.

Fahad090NP:
I am really very happy to see this result. Now I think I should remove the `Draft`. Even if someone want to improve it somehow, tell me I will do the changes.

proof

Fahad090NP:
Hi @ltmoerdani — the branch is now merged with the latest main (incl. #132, #133, #135 — all conflicts resolved, 204/204 tests, strict lint + tsc green) and the diff is clean.

**Please don't hold back on changes in this PR.** I explicitly want to know **each and everything** you'd fix or improve — architecture, naming, UX, settings, edge cases, even typos. I really appreciate the detailed reviews you've given so far, and this feature (inline completions) benefits most from a thorough pass since it's a whole new surface (provider + engine + settings).

A few known points for your review consideration:
- The completion engine is chat-completions-based (no FIM endpoint exists on OpenCode); Qwen `enable_thinking=false` was chosen from live measurements (~1.5s TTFB, zero hidden reasoning).
- The provider competes with Copilot's ghost text — we documented the disable settings.
- Context window is deliberately tiny (10 lines + short suffix); debounce 300ms / timeout 3s / keystroke abort.
- `inlineSuggestions` is opt-in (default `false`).

ltmoerdani:
This is a strong first pass @Fahad090NP. I pulled the branch and ran the full suite and lint locally: 207/207 tests pass, tsc and eslint strict are both clean. The foundation is solid.

The engine decision is the part I like most. Measuring TTFB and hidden reasoning per model instead of guessing, then settling on qwen3.5-plus with enable_thinking=false, is exactly how I'd have approached it. The strict cancellation (300ms debounce, abort on the next keystroke, 3s timeout) and the .vscodeignore catch also stand out. Small detail, big impact: without removing that ignore rule, the feature would silently disappear from the packaged VSIX.

A few things I'd want to look at before this is ready for merge:

1. resolveApiKey only reads the global secret (`src/extension.ts:1008`). Users who set their Go key through a BYOK group, or run multiple profiles, will get nothing and won't know why. The chat path already has a richer fallback (profile key first, then the secret). I'd reuse that, or at minimum document the limitation so it doesn't read as a bug later.

2. The engine swallows failures quietly. Non-200, network errors, and timeouts all return `{ text: undefined }` without writing anything to the log (`src/autocomplete/engine.ts`). The output channel only logs on success, which makes it less useful for the "report what you see" flow you set up for testers. A one-line log with the status or the error reason would turn silent failures into reportable ones.

3. These requests bypass the usage tracker. `tracker.record()` only runs inside the chat provider path, and autocomplete uses a raw fetch, so none of this usage shows up in the status bar or the usage panel. Since Go is paid, someone typing all day could accumulate spend that never appears in the UI. At minimum a note in the docs; ideally a follow-up to wire it in.

4. The debounce knob is read once. `provider.ts` builds the Debouncer in the constructor, so changing `inlineSuggestionsDebounceMs` needs a reload, while the other four knobs are read per request. Minor inconsistency with the config-driven framing.

5. `cleanCompletion` strips leading whitespace (`engine.ts:129`). For a completion that starts at the beginning of a nested line, that can drop the indentation the model produced. Probably fine for mid-line cursors, but a quick test for a nested-block case would settle it.

Small stuff:
- The PR body still says draft, even though you already took it out of draft. Just a cleanup on the description.
- The body says 12 unit tests, the file actually has 15. No big deal, easy to fix.
- `qwen3.7-plus` is in the settings enum and has pricing in the tracker, but it's missing from the fallback model list in `extension.ts` and the metadata snapshot. It'll appear once the live fetch runs, but adding it to the bundled fallback keeps offline mode consistent.
- The output channel gets created even when the feature is off, and the provider registers on `"**"` for every file type. Both harmless since `isEnabled()` gates them, just slightly wasteful.

None of this blocks the idea. Points 1 through 3 are the ones I'd genuinely want answered, or turned into explicit TODOs, before shipping. Happy to talk through any of them if something isn't clear.

Fahad090NP:
Thank You @ltmoerdani, please let me do some more changes in both PRs. There are some bugs that I found. Let me complete everything then I will flag when I am ready. You can review both of the PRs and tell me about the things you found.

Fahad090NP:
Thank you for the detailed review @ltmoerdani — genuinely appreciated. I've worked through every point:

**1. `resolveApiKey` only reads the global secret — fixed.** `registerInlineCompletions` now resolves the key with the same order as the chat path: the active profile's own key first (`profileApiKeys`), then the secret — so multi-profile and BYOK-group users get completions too. Commit `5da9b6c`.

**2. Engine swallows failures — fixed.** The engine now logs every failure to the **OpenCode Completions** output channel with the reason: network error (message), non-OK status (HTTP code + status text), and mid-stream interruptions. The "report what you see" flow now works for failures too. Commit `5da9b6c`.

**3. Autocomplete bypasses the usage tracker — acknowledged, partially addressed.** The per-day **Suggested / Approved** counters (in #138) track how many completions were shown and accepted, persisted daily and charted in the usage panel — but the *cost* of those requests isn't wired into `tracker.record()` yet. That needs the completion engine to emit a transport summary (it's a raw fetch today). I'd like to do it as a follow-up after this merges, and I've noted it in both PR bodies so it doesn't read as an oversight.

**4. Debounce knob read once — fixed earlier.** `Debouncer.delayMs` is now live: a pending run is rescheduled when `inlineSuggestionsDebounceMs` changes, so the setting applies on the very next keystroke (commit `68c3eb6`). Covered by a unit test.

**5. `cleanCompletion` stripping leading whitespace — fixed.** It now strips leading spaces/tabs (the cursor already sits on indented code) but **preserves leading newlines**, so a completion that continues on a nested line keeps its line break and indentation. New unit test covers the nested-block case. Commit `5da9b6c`.

**Small stuff:**
- PR body rewritten — draft wording gone, test counts corrected (15 tests, 207 at merge base).
- `qwen3.7-plus` added to the Go fallback model list and the bundled metadata snapshot (commit `1c26fd1`), matching the settings enum and the tracker's pricing table.
- Output channel created eagerly + `"**"` registration: acknowledged, harmless since `isEnabled()` gates everything — left as-is deliberately.

**New:** I also created issues for every bug fixed in #138 (#139–#148) and linked them in that PR's body.

Happy to adjust anything else you spot.

Fahad090NP:
Follow-up on point 3: I've also added the docs note you asked for — the CHANGELOG now explicitly lists autocomplete cost attribution as a known limitation with the planned follow-up (commit `abbfcd2`), alongside the Suggested/Approved per-day counters that do ship in #138. All five review points are now fully addressed (1–2, 4–5 fixed in code, 3 documented).

ltmoerdani:
nice one on the write-up, this answers everything cleanly. I read through #138 as well, so I have the full picture now.

Merge order stays as you set it up: #136 first, then #138. I'll merge #136 with a merge commit and let it retarget to main. One thing I want on the record so nobody's surprised later: the review fixes live in #138, not in #136. So #136 lands as-is, and #138 carries the key fallback, the failure logging, the indentation fix, and the qwen3.7-plus fallback along with the refactor and the dashboard. That's fine, just making sure we're aligned.

The Suggested/Approved counters and the changelog note are the right way to handle point 3 for now. Wiring the cost into tracker.record() can wait until after this merges, like you said.

Take your time on the rest. I'll start going through #138 properly once #136 is in.

Issue #78, opened 2026-07-17
[BUG] Could not fetch OpenCode Zen (Agents) model list. Using bundled model list. fetch failed


leiyu1980:
## 🐛 What happened?
vscode 1.129.0 After startup, the model list only shows the default model, which is inconsistent with the output of opencode zen list. Additionally, the full list of Zen models sometimes flashes briefly before disappearing.

## 🔄 Steps to reproduce

1.It appears immediately upon startup.

## ✅ What did you expect?

Can pull and display the list normally

## 🖥️ Environment

- **Extension version:** 0.4.1
- **VS Code version:** 1.129.0
- **Provider:** 
- **Model:** All

## 📸 Logs or screenshots (optional)

Could not fetch OpenCode Go model list. Using bundled model list. fetch failed
Could not fetch OpenCode Zen (Agents) model list. Using bundled model list. fetch failed

ltmoerdani:
Thanks @leiyu1980. The "flashes briefly before disappearing" detail is what unblocked this, actually. I'd been reading #78 as a picker crash again, like the one we fixed in #53 last month. Wrong instinct.

Your symptom is something dumber and more boring: the network fetch at startup was fragile, and we never built it to fail gracefully.

Here's what I found in the code. Node's built-in `fetch` is undici under the hood, and undici's defaults are not friendly to a picker that needs to feel snappy. Headers can wait up to 300 seconds before timing out, and there's no connect timeout at all, so one flaky DNS lookup or a stale keep-alive socket at boot could either hang the picker for the full five minutes or silently drop it to the bundled list. VS Code 1.129's new agent host also spins up a lot more concurrent resolution calls into the provider than 1.128 did, which made the race surface more often. The undici maintainers consider this expected behavior, by the way. [nodejs/undici#5450](https://github.com/nodejs/undici/issues/5450) is essentially the same bug report, closed with "use retries."

The shame is that the fix is obvious in hindsight. I should have written it this way the first time.

What changed in #80:

[...]

PR #138, opened 2026-08-13
refactor: central config, shared utils, and verified bug fixes

### LINKED ISSUE #139: kimi-k2.7-code requests 400 on cold start: bundled temperature:false fallback ignored
## Symptom
`kimi-k2.7-code` requests intermittently fail with `HTTP 400` ("invalid temperature: only 1 is allowed") on cold start or when the `models.dev` fetch fails.

## Root cause
`resolveModelMetadata()` resolved `temperature` only from live/cached metadata and never consulted the bundled `temperature: false` fallback (which exists precisely so the parameter is omitted offline). Every cold-start request sent `temperature: 0.2` and depended on the runtime 400-retry to rescue it.

## Impact
Unreliable responses from kimi-k2.7-code when the metadata cache is missing/stale; extra latency from the retry.

## Fix
PR #138 — the fallback chain now includes the bundled value, so `temperature` is omitted for kimi-k2.7-code on every path. Regression-tested.

### LINKED ISSUE #140: Audio/PDF/video models advertise image input they don't support
## Symptom
Models whose input modalities include audio, PDF or video but NOT images advertise `imageInput: true` in the model picker; VS Code then forwards image attachments the model cannot process, producing upstream 400s, and the vision proxy is skipped for them.

## Root cause
Modality detection treated *any* non-text input modality as vision: `inputModalities.some((m) => m !== "text")` → true for `["audio","text"]`, `["pdf","text"]`, `["video","text"]`.

## Impact
Broken image attachments and 400 errors for audio/PDF/video-capable models.

## Fix
PR #138 — vision is now only implied by actual image modalities (`image`, `image_url`, `input_image`); audio/PDF/video flags are tracked separately. Regression-tested.

### LINKED ISSUE #141: freeOnly filter bypassed on cached/bundled model lists (paid Zen models shown offline)
## Symptom
With `opencodego.freeOnly: true`, paid Zen models (e.g. `claude-opus-4-7`, `gpt-5.5`, `qwen3.6-plus`) can appear in the model picker whenever the live model-list fetch fails or is cancelled (offline/degraded mode).

## Root cause
The Zen `freeOnly` filter was applied only on the live model-list fetch path; the cached/bundled fallback lists were returned unfiltered.

## Impact
A free-only user could unknowingly pick (and pay for) a paid model.

## Fix
PR #138 — `definition.filterModel` is now applied on every path: live fetch, cached snapshot, and bundled fallback, including the metadata-success branch of `filterAvailableModels` (flagged in the #138 review). Regression-covered.

### LINKED ISSUE #142: Kimi K2.7 / MiniMax M3 chain-of-thought leaks into visible chat text
## Symptom
Kimi K2.7-code and MiniMax M3 emit their genuine chain-of-thought as **visible chat text** (and it gets echoed as `reasoning_content` on the next turn).

## Root cause
The Go-gateway "thinking-off" workaround (#37635) decides whether `reasoning_content` is genuine CoT by checking only `reasoning_effort`/`budget_tokens`. Kimi K2.7-code (`thinking: { type: "enabled", keep: "all" }`) and MiniMax M3 (`thinking: { type: "adaptive" }`) route through chat-completions with Anthropic-style thinking blocks, which the detector did not recognize — so genuine reasoning was misclassified as "thinking off" and surfaced as visible text.

## Impact
Polluted responses with private reasoning in the visible output.

## Fix
PR #138 — detection now covers every reasoning channel via `bodyRequestsThinking()` (reasoning_effort, budget_tokens, enable_thinking, thinking block types). Regression-tested.

### LINKED ISSUE #143: Re-editing monthly spent targets inflates the baseline (usage card shows wrong spend)
## Symptom
Re-editing the monthly spent target (Set spent targets) inflates the displayed spend. Example: target `$60` with `$12` raw SQLite spend produced a `$72` baseline, so the usage card showed the wrong amount after every re-edit.

## Root cause
`setManualSpentTargets()` subtracted the active baseline from the raw-SQLite monthly sum — but raw SQLite rows never contain baseline amounts (the subtraction is only valid for the summary-based fallback, where `summary.monthly.spent` already includes the baseline).

## Impact
Wrong monthly spend display and incorrect baseline after re-editing targets.

## Fix
PR #138 — the baseline is only subtracted on the summary path; the SQLite path sums raw rows as-is.

### LINKED ISSUE #144: Today/Yesterday/Codebase usage shows 0 when sqlite3 isn't on the extension host PATH
## Symptom
Today / Yesterday / Codebase usage silently shows 0 (status bar, tooltip and usage panel) even though the OpenCode CLI history has data — while the fetched Session/Weekly/Monthly quota keeps working. Flaky across launches: works when VS Code is launched from a shell whose PATH has `sqlite3`, zeros otherwise.

## Root cause
The CLI history was read by spawning the `sqlite3` binary. On many dev machines `sqlite3` resolves to the Android SDK's copy (`~/Android/Sdk/platform-tools/sqlite3`), which is on the PATH only for terminal-launched windows — desktop-launched extension hosts get `ENOENT` and the read silently returned nothing.

## Impact
Completely blank device-level usage for the most common launch methods.

## Fix
PR #138 — the history is now read through Node's built-in `node:sqlite` (`DatabaseSync`, read-only, retried on busy WAL states) with zero external binary dependency; the `sqlite3` binary remains as fallback, resolved from PATH plus known absolute locations (system, Homebrew, Android SDK). All failures are logged with the exact error to the "OpenCode Go Usage" output channel, and activation logs a startup availability line.

### LINKED ISSUE #145: Anthropic stream extractor writes usage fields onto itself instead of the request summary
## Symptom
The Anthropic stream extractor's `message_delta` handler wrote usage fields onto the extractor instance itself (`this as unknown as RequestUsageSummary`) instead of the request summary.

## Root cause
The real usage summary is already updated by the SSE `onData` callback in `parseServerSentEvent`; the duplicate write mutated the wrong object with latent impact on usage reporting.

## Impact
Latent: wrong-object mutation that could corrupt usage accounting if the shared update path ever changes.

## Fix
PR #138 — the duplicate (and wrong-object) write is removed; `message_delta` only flushes tool calls.

### LINKED ISSUE #146: String temperature/maxTokens/timeout values reach the request body and cause HTTP 400
## Symptom
A misconfigured string value in `settings.json` for `temperature`, `maxTokens`, `maxInputTokens`, `requestTimeoutSeconds` or `streamIdleTimeoutSeconds` was forwarded into the request body unvalidated → upstream HTTP 400 with no clear cause.

## Impact
Mysterious request failures after a config typo.

## Fix
PR #138 — all numeric settings are sanitized/clamped with fallbacks (`toFiniteNumber`), and the inline-completion numeric knobs are clamped to their documented ranges.

### LINKED ISSUE #147: Usage 'resets in' shows '26h' instead of '1d 2h'
## Symptom
The usage tooltip/card rendered a 26+ hour reset countdown as `"26h"` instead of `"1d 2h"`.

## Root cause
`formatRelativeTime`'s days branch was checked after the `minutes === 0` shortcut, so multi-day windows fell through to the hours formatting.

## Fix
PR #138 — the days branch now takes precedence; covered by unit tests.

### LINKED ISSUE #148: Error classification can crash on hosts without a global DOMException
## Symptom
Error classification (`isTransientFetchError`, model-list abort check) used `error instanceof DOMException` unguarded — on a host without a global `DOMException`, the error-handling path itself throws a `ReferenceError`, masking the original error.

## Impact
Crash while handling errors on older runtimes; original failure hidden.

## Fix
PR #138 — both sites guard `typeof DOMException === "function"` before `instanceof`.

Fahad090NP:
**Refactor + usage dashboard. Merge #136 first** — this branch is stacked on the autocomplete feature; GitHub auto-retargets to `main` once #136 lands. The 5 autocomplete commits shown here are exactly #136's.

## What

Two intentional halves:

### 1. Central config + shared utilities (behavior-preserving)
- **`src/config.ts`** — every tunable constant (URLs, timeouts, limits, storage keys, setting keys, defaults) in one dependency-free file; modules re-export the names their callers rely on.
- **`src/utils.ts`** — shared primitives replacing copies across modules: `isRecord` (was 6×), `firstString`, `compactErrorCode`, `positiveNumber`, `toFiniteNumber`, `getErrorMessage`, `formatUsd`, `formatCount`/`formatTokenCount` (K/M/B/T), `formatRelativeTime`, `escapeHtml`, cancellable delays.
- Both stream extractors share a `BaseResponseExtractor`; verification scripts import the real production logic; 4 dead exports removed.

### 2. Verified bugs fixed (issues #139–#148)
- **#139** kimi-k2.7-code 400 on cold start (bundled `temperature:false` fallback ignored)
- **#140** audio/PDF/video models advertise image input they don't support
- **#141** `freeOnly` filter bypassed on cached/bundled model lists
- **#142** Kimi K2.7 / MiniMax M3 chain-of-thought leaked into visible text
- **#143** re-editing monthly spent targets inflated the baseline
- **#144** Today/Yesterday/Codebase showed 0 when `sqlite3` wasn't on the host PATH (now read via `node:sqlite`)
- **#145** Anthropic stream extractor mutated itself instead of the usage summary
- **#146** string settings values caused HTTP 400 (numeric sanitization)
- **#147** `26h` instead of `1d 2h` in the usage reset countdown
- **#148** error classification could crash without a global `DOMException`

### 3. Usage — real data, realtime, full dashboard
- **Today/Yesterday merge the OpenCode CLI history** (per-message cost, tokens, `path.cwd`) with extension-tracked requests; source selectable (`usageTodayYesterdaySource`).
- **Codebase** row (all-time usage in the current workspace), permanent tracking (no reset action, no 31-day cutoff), instant startup (persisted server snapshot), 3s memoized CLI reads, background refresh loop + manual Refresh.
- **Full-page panel**: animated ring meters, Spend/Requests/Tokens line charts, Models tab (overlapping per-model lines), Suggested/Approved tabs (per-day completion counters), Window toggle (Week → 14 days → Month → Lifetime, lifetime default), whole-chart hover tooltips, round axis steps.
- Compact K/M/B/T formatting everywhere.
- New settings: `usageTodayYesterdaySource`, `usageCodebaseRow`, `usageCodebaseWindowDays`, `usageRollingSessionMeter`, `usageDayBoundary`, `usageRefreshIntervalSeconds`, `usageChartDays`.

## Review notes (addressed)

- `freeOnly` now applies on every path, including the metadata-success branch of `filterAvailableModels` (commit `d1c9c59`).
- Model names in the panel legend/tooltips are HTML-escaped (commit `d1c9c59`).
- `resolveApiKey` for completions uses the profile-first fallback; engine failures are logged; `cleanCompletion` keeps leading newlines (commit `5da9b6c`).

## Verification

276 unit tests, strict ESLint + tsc + prettier + markdownlint green, VSIX packages and installs cleanly.

---

Fixes #139, Fixes #140, Fixes #141, Fixes #142, Fixes #143, Fixes #144, Fixes #145, Fixes #146, Fixes #147, Fixes #148


Fahad090NP:
**Please merge PR #136 first.** This PR is stacked on the autocomplete feature (#136): it shows the same 5 autocomplete commits plus 7 refactor commits. Once #136 is merged, this PR auto-retargets to `main` and only the refactor commits remain. The refactor touches the autocomplete module (config re-exports), so the merge order matters.

Fahad090NP:
## Bugs users may face in production — fixed by this PR

This PR went through a full-codebase audit that uncovered several real, user-facing bugs that were previously unreported. Here is the complete list, framed by what users would actually experience:

### 1. `kimi-k2.7-code` requests fail on cold start / offline
`resolveModelMetadata()` never consulted the bundled `temperature: false` fallback — the request body still included `temperature: 0.2`, which Moonshot rejects with `HTTP 400` (only `1` is allowed). Every cold-start request (failed `models.dev` fetch) relied on the runtime 400-retry to rescue it. **Now:** the fallback chain includes the bundled value, so the parameter is omitted.

### 2. Audio/PDF/video-capable models advertise image support they don't have
Vision detection treated *any* non-text input modality as vision (`input.includes("image") || "audio" || "pdf"` → true). Models with `modalities: ["audio","text"]` or `["pdf","text"]` reported `imageInput: true`, so VS Code forwarded image attachments the model cannot process → upstream `400` errors, and the vision proxy was skipped for them. **Now:** only actual image modalities imply vision.

### 3. `freeOnly` users were offered paid models in degraded mode
The Zen `freeOnly` filter was applied only on the live model-list fetch. When the fetch failed or was cancelled (network hiccup, hung TCP — issue #78 territory), the cached/bundled fallback list was returned **unfiltered**, showing paid models (`claude-opus-4-7`, `gpt-5.5`, `qwen3.6-plus`, …) in the picker. A free-only user could unknowingly pick a paid model. **Now:** the filter applies on every path (live, cache, bundled).

### 4. Kimi K2.7 / MiniMax M3 thinking leaked into visible chat text
The Go-gateway "thinking-off workaround" (#37635) decided whether `reasoning_content` was genuine CoT by checking only `reasoning_effort`/`budget_tokens`. Kimi K2.7-code (`thinking: { type: "enabled", keep: "all" }`) and MiniMax M3 (`thinking: { type: "adaptive" }`) route through chat-completions with Anthropic-style blocks — neither was detected, so **genuine chain-of-thought was emitted as visible text** (and echoed as `reasoning_content` on the next turn via the reasoning marker). **Now:** detection covers every reasoning channel (`bodyRequestsThinking()`).

### 5. Re-editing monthly spent targets inflated the baseline
`setManualSpentTargets()` subtracted the active baseline from the raw-SQLite monthly sum — but raw SQLite rows never contain baseline amounts. Example: target `$60` with `$12` raw spend produced a `$72` baseline, so the usage card showed the wrong spend after every re-edit. **Now:** the baseline is only subtracted on the summary path (where it is actually included).

### 6. SQLite-backed usage summaries reported `sqliteAvailable: false`
The rows-based summary builder returned the inverted flag (only masked by the enriched builder overriding it afterwards) — duplicated logic that could diverge. Cosmetic on its own, but wrong diagnostic output in the usage card.

### 7. Anthropic stream extractor wrote usage fields onto itself
The `message_delta` handler mutated the extractor instance (`this as unknown as RequestUsageSummary`) instead of the request summary — the real summary was already updated by the SSE data callback. Wrong-object mutation with latent impact on usage reporting; the duplicate write is now removed.

### 8. Misconfigured numeric settings could 400 every request
A string `temperature`, `maxTokens`, or timeout value in `settings.json` was forwarded into the request body unvalidated → upstream HTTP 400 with no clear cause. **Now:** all numeric settings are sanitized/clamped with fallbacks.

### 9. `inlineSuggestionsDebounceMs` changes were ignored until restart
The debouncer captured its delay at construction; config changes never rescheduled a pending run. **Now:** `Debouncer.delayMs` is live — a pending run is rescheduled, so the setting applies on the very next keystroke.

### 10. Usage "resets in" display bug
`formatRelativeTime` rendered 26+ hours as `"26h"` instead of `"1d 2h"` (days branch lost to the minutes-shortcut) — wrong reset countdown in the usage tooltip/card.

### 11. Error classification could crash while handling errors
`isTransientFetchError` / the model-list abort check used `error instanceof DOMException` unguarded — on a host without a global `DOMException` (older runtimes), the error-handling path itself would throw a `ReferenceError`, masking the original error.

### 12. Zen picker duplication guard relied on a useless ternary
`sanitizeToolSchema` emitted `type: "object"` through a ternary whose branches were identical (dead code, not user-visible, but a trap for future edits).

---

**In addition** (no user-visible behavior change): `src/config.ts` centralizes every tunable constant, `src/utils.ts` replaces 6× `isRecord` and other copy-pasted helpers, both stream extractors share a `BaseResponseExtractor`, the verification scripts now import the real production logic instead of drifted copies, and 4 dead exports were removed. Test suite grew from 207 → 246 with regression tests for items 1, 2, and 4.

Fahad090NP:
From there:
Pasted image
To Here:
Pasted image (2)

Fahad090NP:
## Complete summary of everything in this PR

> Stacked on #136 (merge that first — these commits target the refactor + usage work; GitHub auto-retargets to `main` after #136 lands).

### 1. Central configuration — `src/config.ts`
Every tunable value (URLs, timeouts, limits, storage keys, setting keys, defaults, GO subscription tiers) now lives in one dependency-free file. Modules import from it and re-export the names their callers/tests rely on — nothing external changed.

### 2. Shared utilities — `src/utils.ts`
Replaced near-identical copies scattered across modules: `isRecord` (was defined 6×), `firstString`, `compactErrorCode`, `positiveNumber`, `toFiniteNumber`, `getErrorMessage`, `formatUsd`, `formatCount`/`formatTokenCount` (K/M/B/T with unit escalation), `formatRelativeTime`, `escapeHtml`, and the two cancellable delay helpers.

### 3. Verified production bugs fixed
- **kimi-k2.7-code 400s on cold start** — `resolveModelMetadata` ignored the bundled `temperature: false` fallback.
- **Audio/PDF/video models advertised image input** — vision detection now requires actual image modalities.
- **Zen `freeOnly` filter bypassed** on cached/bundled model lists (paid models shown to free-only users offline).
- **Kimi K2.7 / MiniMax M3 chain-of-thought leaked into visible text** — the Go-gateway thinking-off workaround now recognizes Anthropic-style `thinking` blocks.
- **`setManualSpentTargets` baseline double-count** inflated re-edited monthly targets.
- **Anthropic stream extractor mutated itself instead of the usage summary**; inverted `sqliteAvailable`; useless ternary in `sanitizeToolSchema`; duplicated Audio badge; `formatRelativeTime` 26h→`26h`; DOMException crash risk in error classification.
- Both stream extractors now share a `BaseResponseExtractor`; verification scripts import the real production logic instead of drifted copies; 4 dead exports removed.

### 4. Usage — real data, realtime, beautiful
- **Today/Yesterday merge the OpenCode CLI history with extension-tracked requests** (per-message `cost` + `tokens.{input,output,reasoning,cache}` + `path.cwd` from `opencode.db`). Source selectable: `usageTodayYesterdaySource` (auto/cli/extension).
- **Codebase row replaced "Session (est)"** and is renamed **Total spend** everywhere — all-time usage in the current workspace (per-directory via `path.cwd`), windowed by `usageCodebaseWindowDays`.
- **Tracked data is permanent** — the "Reset tracked usage data" action and 31-day entry cutoff are gone.
- **Instant startup + fast reads** — the server snapshot is persisted and restored (`opencodego.serverUsage.v1`), and the multi-GB CLI read is memoized (3s TTL).
- **Realtime** — a background refresh loop (`usageRefreshIntervalSeconds`, default 60s), immediate repaint on config changes, and a manual Refresh button.
- **Compact formatting** — `1.2M` tokens, `1.7k` requests, `$1.23M` / `$0.0004`-style sub-cent dollars — everywhere.

### 5. Full-page usage dashboard (webview)
Animated ring meters (Session/Weekly/Monthly), stat chips, tabs:
- **Spend / Requests / Tokens** — line charts with whole-chart hover (nearest day: guide line + dots + cursor-follow tooltip).
- **Models** — overlapping colored line charts per model (spend/day), ranked legend, per-day tooltips listing every model's spend.
- **Suggested / Approved** — inline chat-completion counters (suggestions shown by the provider; acceptances detected via a bounded insert heuristic — VS Code has no acceptance API, verified in `vscode.d.ts`). Whole-number axes, honest tooltips, same day buckets as the other charts (no undefined hovers).
- **Window button** (Week → 14 days → Month → Lifetime, lifetime default), Set targets / Rename / Refresh buttons in the top-right.
- Round, equal tick steps on every axis (1/2/2.5/5 × 10ⁿ).

### 6. Settings added
`usageTodayYesterdaySource`, `usageCodebaseRow`, `usageCodebaseWindowDays`, `usageRollingSessionMeter`, `usageDayBoundary`, `usageRefreshIntervalSeconds`, `usageChartDays`.

### Verification
275 unit tests, full lint (ESLint + tsc + prettier + markdownlint) green, VSIX packages and installs cleanly.

Fahad090NP:
From this:
Pasted image
To this:
Pasted image (2)

ltmoerdani:
Nice work on this @Fahad090NP. The refactor is genuinely clean to read: `config.ts` and `utils.ts` being dependency-free makes the whole thing easier to reason about, and I verified the bug fixes one by one. Pulled the branch, compiled, ran the suite: 275 green, zero failures.

One thing I want to flag before we merge, and it's more of a scope question than a code question.

The PR body says once #136 lands, only the 7 refactor commits remain. But the branch actually carries 15 commits on top of that: a whole usage dashboard feature. Full-page panel with rings and charts, compact K/M/B/T formatting, permanent tracking, the today/yesterday/codebase rows. It's not mentioned in the description, and it's not on main yet. So after #136 merges, this PR still pulls those 15 in along with the refactor.

Was that intentional? If the dashboard was meant to go out separately, we can carve it out. If it's part of this PR, that's fine too, I just want the changelog and the description to reflect reality before it ships. Merge order stays the same either way: #136 first, then this.

Two smaller things I noticed while reading, both low priority:

1. `freeOnly` filter. The fix covers the live fetch path and the error path of `filterAvailableModels`, but the success path still skips `filterModel`. So in the narrow case where the model-list fetch fails but the models.dev metadata fetch succeeds, the bundled fallback could still show paid Zen models (`kimi-k2.5`, `qwen3.6-plus`) to free-only users. Cached lists are fine, they're filtered at fetch time. Probably worth a follow-up rather than blocking this.

2. In the usage panel, model names land in the tooltip/legend via `innerHTML` without `escapeHtml`. Data comes from the gateway, so real-world risk is low. Still, a quick escape is cheap insurance.

Happy to merge once we settle the scope question. Nice work either way.

Fahad090NP:
Thank you @ltmoerdani — and great catches on both points. Addressed:

**Scope question — yes, intentional.** The usage dashboard was always meant to ride along with this PR (the usage rework depends on the refactor's `config.ts`/`utils.ts` anyway). I've rewritten the PR body to reflect reality: central config + shared utils, the ten verified bugs fixed (#139–#148 — I created issues for each and linked them in the body), and the full usage dashboard (Today/Yesterday/Codebase rows, permanent tracking, realtime loop, the panel with rings/line charts/Models/Suggested/Approved tabs and the window toggle). Merge order stays: #136 first, then this.

**1. `freeOnly` on the metadata-success path — fixed.** You were right: the fix covered the live-fetch path and the error path of `filterAvailableModels`, but the success path skipped `filterModel`. `definition.filterModel` is now applied in **every** branch — live fetch, cached snapshot, bundled fallback, and both metadata paths of `filterAvailableModels` (commit `d1c9c59`).

**2. Unescaped model names in the panel — fixed.** The legend and hover tooltips now HTML-escape model names before `innerHTML` insertion (`esc()` in the webview), so a gateway-supplied model ID can't inject markup (commit `d1c9c59`).

**Also worth mentioning:** during the same pass I added `qwen3.7-plus` to the bundled fallback lists (commit `1c26fd1`) — flagged in the #136 review — and the autocomplete key resolution / failure logging / indentation fixes from that review are in this branch too (commit `5da9b6c`).

Verification stands at **276 unit tests**, strict lint + tsc + prettier + markdownlint green, VSIX packages and installs cleanly. Ready for merge after #136.

ltmoerdani:
Checked the latest branch again. Compiled clean, 276 tests green. I also went straight into the code for the two fixes instead of trusting the commit messages.

`freeOnly` is applied in every branch of `filterAvailableModels` now, including the metadata-success path you flagged. And the webview escapes model names in both the legend and the hover tooltips via `esc()`. Both read correct.

The `node:sqlite` fallback for #144 is a nice catch. Depending on the `sqlite3` binary being on the host PATH was always fragile, and the Today/Yesterday/Codebase rows silently zeroing out is exactly the kind of failure nobody notices until a user complains.

#136 is merged now, so this one is next in line. Everything checks out on my end, ready to merge whenever you are.

Nice work on the whole thing. The refactor and the dashboard both landed clean.

PR #150, opened 2026-08-14
fix(streaming): DeepSeek V4 reasoning goes to the thinking block, never visible text

Fahad090NP:
## What

DeepSeek V4 keeps producing genuine chain-of-thought even when the request carries no `reasoning_effort` — "thinking off" only omits the effort parameter, it does not stop the model from reasoning. The Go-gateway "thinking-off" workaround (#37635) classified any `reasoning_content` as *mislabeled content* when thinking was off, so DeepSeek's real CoT was printed as **visible chat text**.

## Fix

`treatReasoningAsContent` is now model-aware:

- It only applies to models that genuinely stop reasoning when no effort is sent — the verified case is MiMo (where the gateway wraps the actual response into `reasoning_content`).
- It never applies to reasoning-first families: `modelAlwaysReasons()` (currently the `deepseek-*` family). Their `reasoning_content` always lands in the **thinking block** (hidden/collapsed per `chat.agent.thinkingStyle`).

## Also removed (dead branches in the same path)

- The redundant `isGoGateway &&` wrapper inside `hasReasoningEffort` — only consulted under the `isGoGateway` guard and re-checked by `treatReasoningAsContent`.
- The implied-empty `!visible &&` in the reasoning-emission condition — `filterText("")` always yields an empty visible slice.

## Verification

Verified live by the author with `deepseek-v4-flash` (thinking off): reasoning now appears in the thinking block, the visible response is clean. 276 unit tests, strict ESLint + tsc + prettier + markdownlint green, VSIX packages and installs cleanly.

## Notes

`[go-gw]` output-channel lines now also show the model-aware decision for debugging.


ltmoerdani:
Thanks @Fahad090NP, this one's clean. I ran the whole thing locally: compile passes, 276/276 unit tests green, ESLint + tsc + prettier + markdownlint all clean. Branch is based on current main, so no conflicts.

The family-aware split is the right fix. DeepSeek still reasons when thinking is off, MiMo actually stops, and the old code treated both the same way. That's exactly how CoT ends up leaking into visible text. Keeping the #37635 workaround scoped instead of gutting it was the safer move.

One tiny nit, non-blocking: the JSDoc on the `treatReasoningAsContent` constructor param still reads "i.e. MiMo thinking is OFF", which is now slightly stale since the flag is also off for the whole deepseek family. One-line touch-up if you feel like it.

Good to merge from my side.

PR #154, opened 2026-08-14
fix(autocomplete): no ghost text in the Copilot Chat prompt box

### LINKED ISSUE #153: Inline suggestions appear in the Copilot Chat prompt box while typing
## Symptom
With `opencodego.inlineSuggestions` enabled, ghost-text suggestions also appear in the **GitHub Copilot Chat prompt box** while typing a prompt — completions belong in code editors, not in chat prompts.

## Root cause
The inline-completion provider registers on the document selector `"**"` (every document). VS Code exposes the chat prompt as a virtual editor document with the `chatSessionInput` / `sessions-chat` schemes (`chatInputSchemes` in VS Code's `workbench/contrib/chat/common/constants.ts`), so the provider is asked for suggestions there too.

## Fix
PR — the provider early-returns for chat-input documents via the pure `isChatInputDocument()` helper (checks the two chat-input schemes), keeping ghost text in code editors only. Unit-tested (278 tests).

Fahad090NP:
Fixes #153 — implemented as an **opt-in feature**.

## What

With `opencodego.inlineSuggestions` enabled, ghost-text suggestions also appeared in the **GitHub Copilot Chat prompt box** while writing prompts — VS Code exposes the prompt as a virtual `chatSessionInput` / `sessions-chat` document (`chatInputSchemes` in `workbench/contrib/chat/common/constants.ts`) and our `"**"` selector matched it.

Instead of a hard block, this ships as a user choice:

- **New setting `opencodego.inlineSuggestionsChatInput`** (default `false`) — when on, completions are also offered while typing prompts; off keeps ghost text in code editors only.
- The provider gates chat-input documents through the pure `isChatInputDocument()` helper (verified schemes), checking the setting before building context or starting any request.

## Verification

Unit tests for `isChatInputDocument` (chat schemes rejected, `file`/`untitled` accepted); 278 tests, strict ESLint + tsc + prettier + markdownlint green. Not installed — awaiting review/merge.


ltmoerdani:
@Fahad090NP clean fix, small scope, CI pass, 278 tests green.

`isChatInputDocument()` pure helper, testable without VS Code API. Opt-in via setting makes sense — default off, backward-compatible.

One note: `CHAT_INPUT_SCHEMES` hardcodes 2 values (`chatSessionInput`, `sessions-chat`). If VS Code adds new schemes in a future version, this needs a manual update. Already documented in the comment, so not a blocker, but worth flagging.

Merge when ready.

Fahad090NP:
Thanks for the review @ltmoerdani — addressed your note about the hardcoded schemes.

The provider now serves only real editable code surfaces via a **code-editor scheme allowlist** (`CODE_EDITOR_SCHEMES`: `file`, `untitled`, `git`, `vscode-userdata`, `vscode-notebook-cell`). Any interactive surface VS Code introduces in the future — new chat-prompt variants, webviews, output, custom editors — is excluded automatically, so no manual updates needed. The opt-in carve-out for the chat prompt (`opencodego.inlineSuggestionsChatInput`) is preserved: chat-input documents are still allowed when the setting is on.

New pure helper `isCompletionDocument()` (alongside `isChatInputDocument()`), both unit-tested — 280 tests, strict lint + tsc green. Ready when you are.

ltmoerdani:
Nice update. The allowlist approach is better than the original blocklist. Future VS Code interactive surfaces are excluded automatically, and genuinely new code surfaces are rare enough that manual updates are fine.

Tests cover the right cases. CI still green. Approved.

PR #155, opened 2026-08-14
refactor: split god files into domain modules + data-driven model registry

xianhongtao:
## 📝 What does this change?

Splits the three god files — `src/extension.ts` (4653 lines), `src/streaming.ts` (1620) and `src/goUsageTracker.ts` (1510) — into domain modules (behavior-preserving), and adds a data-driven model registry so per-model wiring (transport + thinking family) lives in one table.

- **`src/usage/`** — Go usage domain: `tracker.ts`, `history.ts` (OpenCode CLI SQLite read/aggregation), `pricing.ts`, `formatting.ts`, `dashboard.ts` (status bar + usage webview + tooltip SVG), moved `usage.ts` / `usageProfile.ts` / `goUsageSync.ts`.
- **`src/transports/`** — one file per transport (`chatCompletions`, `responses`, `anthropic`, `google`) + shared `engine`, pure `sse` parser, `extractors`, `extract`, `streamParts`, `thinkTags`; contract types in `src/core/transport.ts` (routing in `src/core/routing.ts`).
- **`src/provider/`** — `OpenCodeProvider` class, `definitions` (PROVIDERS + model types), `messages`/`tokens` (conversion + token estimation), `settings`, `visionProxy`.
- **`src/models/`** — `metadata`, `modelLimits`, `modelCapabilities`, `modelNames`, `pricing`, `metadataFetcher` (models.dev cache).
- **`src/commands/`** — provider / agent-window / diagnostics / thinking-picker handlers; `src/request/headers.ts`.
- **`src/core/registry.ts`** — data-driven `MODEL_REGISTRY`: rows map model-family patterns → `{ endpointKind, sdkPackage, thinkingFamily, vendors? }`; `resolveModelRouting()` (transport) and `thinkingFamily()` both read it, so adding a model family = one row (+ optionally a thinking strategy class).
- **`extension.ts`** is now a thin entry (~414 lines) that only wires activation + command registration; the compat barrels (`streaming.ts`, `goUsageTracker.ts`) are removed and every importer references canonical paths.

## 🧪 How did you test it?

- **Live (VS Code Debug host)**: ran the extension via F5 Extension Development Host for an extended session, calling **DeepSeek V4 Flash** through the refactored transport path (streaming, reasoning, tool calls) — no issues observed.
- `npm run compile` — clean
- `npm test` — **305/305** (incl. 14 new `registry.test.ts` cases: transport×vendor routing, thinking family, lookup mechanics)
- `npm run test-retry` — 7/7 (mock-server streaming/retry E2E)
- `npm run lint` — fully green (editorconfig / eslint / markdown / prettier / shell / tsc / tests)
- `npm run package` — VSIX builds (105 files, 2.67 MB)

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works (live DeepSeek V4 Flash via Debug host + unit suite + mock-server E2E)


xianhongtao:
屏幕截图 2026-08-14 170519

ltmoerdani:
Nice one @xianhongtao, solid refactor.

Cross-checked against `docs/architecture/02-20260809-provider-adapter-architecture.md` and all 5 Strangler Fig phases are done. Target folder structure matches 1:1. The registry in `core/registry.ts` is what we had in mind: one row per model family, routing + thinking as data not logic.

A few things before merge:

1. **Merge conflicts:** this branch is CONFLICTING with `main`. Needs a rebase. There are 3 open PRs in the queue (#154, #149, #152) so better to let those land first, then resolve conflicts to avoid going back and forth.

2. **`dashboard.ts` at 1,317 lines:** creeping into god-file territory again. Not a blocker now, but worth splitting later (maybe rings vs tabs as separate modules).

3. **`OpenCodeProvider.ts` at 1,215 lines:** if the provider class keeps growing, split it up.

Scope is big (87 files, +9k/-8k) but everything is behavior-preserving and tests are green (305/305).

Merge priority: #154 → #149 → #152 → then #155.

ltmoerdani:
Hey @xianhongtao, thanks for this. I pulled the branch and ran it locally: compile is clean, 310/310 unit tests pass (a bit more than the 305 you listed, the registry work added a few), test-retry is 7/7 on the mock server, and eslint has nothing to complain about. I also diffed the old routing and thinking logic against the registry table line by line and the behavior lines up: GPT to Responses, Claude to Messages, minimax-m2 split by vendor, Gemini only on Zen, everything else chat-completions. The thinking payloads (deepseek effort, qwen budget, minimax adaptive, the kimi-k2.7 force-on) all match what we had before.

The structure is a clear win. extension.ts went from 4600+ lines to a thin entry point, transports are split per endpoint, usage is in its own domain folder, and adding a model family as one registry row is much easier to reason about than editing routing and thinking in two places. The tests cover the tricky bits too, like the minimax-m2 vs minimax ordering and the vendor restrictions.

One thing I want to flag before we merge, and it's around the API key change. Previously Go and Zen shared the same `opencodego.apiKey` secret, and this PR splits them so Zen now reads `opencodezen.apiKey`. That fixes the collision where one provider overwrote the other's key, which is good. But anyone who set their Zen key through the old Set API Key command has it sitting in `opencodego.apiKey`, and after this change Zen looks in the new slot, finds nothing, and shows zero models until they re-add via BYOK. I don't see a one-time migration in the PR. Two options: copy the legacy key into the Zen slot on activation when it's empty and Go doesn't hold a separate key, or ship this as a documented breaking change with a clear note in the release notes. I lean toward the migration since it's a few lines, but happy to hear your take.

The thinking refactor also drops the globalState shadow copy in favor of a single config authority. That's a behavior change (a good one, it stops one model's thinking effort leaking onto another) rather than a pure refactor, so it deserves a manual spin on a couple of models before release. I don't think it blocks the merge.

One more thing on release timing: given how big this is, I don't plan to ship it right after merging. I want to sit on it for a bit, run it through a longer live session across several models and both vendors, and only cut a release once I'm confident nothing regressed. The merge itself doesn't mean it goes out the door.

Merge-wise I'll keep the full commit history (merge commit, no squash) so all 24 commits stay intact.

PR #156, opened 2026-08-14
docs: add AGENTS.md agent guidelines aligned with repo conventions

Fahad090NP:
## 📝 What does this change?

Adds `AGENTS.md` at the repo root — the **machine-readable agent contract** that GitHub natively surfaces to Copilot and other AI coding agents. It codifies this repo's existing automation rules (never push/PR/merge without permission, feature branches, Conventional Commits, strict-lint verification) and encodes the review standards we've converged on, all aligned with the codebase's actual style rather than a generic template.

Docs-only — no runtime code.

## 🧪 How did you test it?

- `prettier --check` passes on `AGENTS.md`
- `markdownlint-cli2` (repo config) passes — 0 issues
- Pre-commit husky gate (staged-lint + lint-staged) passed on the commit
- CI runs on this PR

## ✅ Checklist

- [x] `npm run compile` passes (docs-only — no code change)
- [x] `npm run lint` passes (docs-only — markdown/prettier verified)
- [x] `npm test` passes (docs-only — no code change)
- [ ] `npm run package` produces a VSIX (N/A — docs-only)
- [ ] I tested it works (N/A — docs-only)
- [x] I updated docs/CHANGELOG if needed (CHANGELOG left for release-time per repo convention)

---

> ✨ *PEACE BE UPON YOU* 🤲

Issue #63, opened 2026-07-01
[BUG] When I have two opencode, the usage panel display error


taojunnan:
always display the first one

expect dynamic switching

taojunnan:
Both adding a new provider and updating an API key will cause the panel statistics to display incorrectly.

ltmoerdani:
Hey @taojunnan , thanks for the report.

Quick question to make sure I understand the setup. When you say "two opencode," do you mean two separate OpenCode Go API keys from different subscription accounts? And you swap between them depending on the project or client?

I've only tested with a single key so far, but looking at the code, I think I see the problem. The Go Usage Tracker reads from a single SQLite database (`opencode.db`) and hardcodes a filter for `providerID = 'opencode-go'`. When you swap API keys, nothing tells the tracker to re-query, so it just keeps showing data from whichever key loaded first.

Two questions to nail down the fix:
1. Are you switching keys via the "Set API Key" command, or through VS Code settings directly?
2. Are both keys Go, or is one Go and one Zen?

That context will point me toward the right approach.

taojunnan:
Hi @ltmoerdani , thanks for the quick response!

Yes, your understanding is exactly correct. I have two separate OpenCode Go API keys from different subscription accounts  

To answer your two questions:

Switching method: I switch keys at `Manage Language Models` click `update api key`

Key types: Both keys are Go accounts. (I don't use Zen at all).

Let me know if you need me to test any beta fixes or provide more logs!

taojunnan:
> Hey [@taojunnan](https://github.com/taojunnan) , thanks for the report.
> 
> Quick question to make sure I understand the setup. When you say "two opencode," do you mean two separate OpenCode Go API keys from different subscription accounts? And you swap between them depending on the project or client?
> 
> I've only tested with a single key so far, but looking at the code, I think I see the problem. The Go Usage Tracker reads from a single SQLite database (`opencode.db`) and hardcodes a filter for `providerID = 'opencode-go'`. When you swap API keys, nothing tells the tracker to re-query, so it just keeps showing data from whichever key loaded first.
> 
> Two questions to nail down the fix:
> 
> 1. Are you switching keys via the "Set API Key" command, or through VS Code settings directly?
> 2. Are both keys Go, or is one Go and one Zen?
> 
> That context will point me toward the right approach.

If you need more opencode go, I can share my API key with you

Wallacy:
[...]

PR #157, opened 2026-08-15
fix: bug-hunt pass — 35 fixes across streaming, usage, providers, models and tooling

Fahad090NP:
## 📝 What does this change?

A **bug-hunt pass over the whole codebase** — 35 small, atomic, one-fix-per-commit changes (32 files, +505/−128). Every fix was verified against the OpenCode gateway source and VS Code source, and the full lint gate (eslint strict + tsc + prettier + markdownlint + shellcheck + all unit tests) is green.

### Highlights (confirmed bugs)

**Crashes / dangerous**
- Webview **Models tab crashed** with `ReferenceError: esc is not defined` — added the missing HTML-escape helper.
- Tool-schema sanitizer could **infinite-recurse on a cyclic schema** and blow the extension-host stack — added a cycle guard (+ tests).

**Streaming**
- 5xx `Router.Unavailable` retry branch was dead (the 5xx body was never read) — now read + classified.
- Cancellation during 5xx backoff surfaced a misleading stale gateway error — now fails cleanly as "cancelled".
- Google tool calls had **empty ids** and per-chunk indices — now get stable synthetic ids.
- Reasoning could be silently dropped when no progress sink existed — now falls back to the legacy emit path.
- Reasoning marker data part was counted as phantom tokens in estimates — now treated as internal data.
- Empty-response diagnostics were missing on the Anthropic/Google transports — added.

**Usage / cost**
- **Day bucketing used `Math.round`** → afternoon events landed in the next day — now `floor` (+ regression test).
- Subscription percent could go **negative** with a baseline over-correction — now clamped to [0,100].
- Unknown/missing models tracked as **$0** offline — now get a conservative fallback estimate.
- sqlite3 history read capped at 64MB → **large histories silently vanished** — raised to 256MB.
- Profile deletion left `serverUsage`/`everTracked` stale keys — now cleared.
- The active profile was being **silently overridden** by provider re-resolution ~every 300ms — a user's explicit choice is now respected.
- `formatUsd` rendered `$-5.00`; `escapeHtml` missed `'` — fixed.

**Provider / models**
- `family` was a per-model unique string, breaking VS Code's family selection/grouping — now a real family name.
- Thinking setting values were unvalidated (a bad `reasoning_effort` → 400) — now validated against the allowed enums; the picker derives from the same allowlist.
- `temperature` and the timeout settings were unclamped — now bounded.
- Tool schemas lost top-level enums and common keywords (`pattern`/`format`/`default`) — preserved (+ tests).
- Model-list sync drift (dead `ring-2.6-1t`, missing `minimax-m3` in the Go fallback) — cleaned up.

**Agents / vision / misc**
- `autoEnableAgentsWindow` off never reverted the auto-enabled core settings — now reverts.
- Autocomplete stale-token race could kill the newest pending suggestion — fixed.
- Vision proxy: empty descriptions now keep a placeholder; whole-conversation mode reuses the cache.
- `reasoning_content` preserved when consecutive assistant messages merge.
- Diagnostics: one failing vendor no longer aborts the whole report; `retry-after` durations are capped.

## 🧪 How did you test it?

- `npm run lint` — all 7 steps green, incl. the full unit suite.
- Pre-commit gate exercised on every one of the 35 commits.
- Verified against `tmp/vscode` + `tmp/opencode` reference sources where behavior depended on VS Code / the gateway.

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [ ] `npm run package` produces a VSIX (not run — code + tests only)
- [ ] I tested it works (live API verification still recommended for streaming/tool-call paths)
- [ ] I updated docs/CHANGELOG if needed (left for release-time per repo convention)

---

> ✨ *PEACE BE UPON YOU* 🤲

### DIFF
--- a/scripts/staged-lint.ts
+++ b/scripts/staged-lint.ts
@@ -49,9 +49,11 @@ function run(cmd: string, args: string[]): CommandResult {
   return { status: res.status, output: `${res.stdout}${res.stderr}`.trim() };
 }
 
-/** Staged (added/copied/modified) file paths relative to the repo root. */
+/** Staged (added/copied/modified/renamed) file paths relative to the repo root. */
 function stagedFiles(): string[] {
-  const res = run("git", ["diff", "--cached", "--name-only", "-z", "--diff-filter=ACM"]);
+  // Include renamed (R) files and enable rename detection so a staged rename's
+  // new path is linted too.
+  const res = run("git", ["diff", "--cached", "--name-only", "-z", "--find-renames", "--diff-filter=ACMR"]);
   if (res.status !== 0) {
     return [];
   }
@@ -88,6 +90,13 @@ function resolveImport(fromFile: string, spec: string): string | undefined {
     path.join(base, "index.ts"),
     path.join(base, "index.js"),
   ];
+  // NodeNext-style: in ESM, `./foo.js` resolves to `./foo.ts`. Without this,
+  // changing foo.ts would not lint its dependents (this repo imports ESM
+  // scripts with `.js` specifiers that map to `.ts` sources).
+  if (/\.(js|cjs|mjs)$/.test(base)) {
+    const tsBase = base.replace(/\.(js|cjs|mjs)$/, "");
+    candidates.unshift(`${tsBase}.ts`, `${tsBase}.tsx`);
+  }
   for (const candidate of candidates) {
     try {
       statSync(candidate);
--- a/src/autocomplete/provider.ts
+++ b/src/autocomplete/provider.ts
@@ -89,7 +89,12 @@ export class OpenCodeInlineCompletionProvider implements vscode.InlineCompletion
       };
 
       const tokenSubscription = token.onCancellationRequested(() => {
-        this.debouncer.cancel();
+        // Do NOT cancel the shared debouncer here: VS Code may cancel this
+        // request's token AFTER a newer keystroke already scheduled its own
+        // debounced run, and aborting the debouncer would kill that newer
+        // pending suggestion. The debouncer cancels the previous run itself
+        // when the next debounce() is scheduled; here we only resolve this
+        // request's promise as "no suggestion".
         finish(undefined);
       });
 
--- a/src/chatParts.ts
+++ b/src/chatParts.ts
@@ -23,7 +23,11 @@ export function createUsageDataParts(usage: UsageSnapshot): vscode.LanguageModel
 }
 
 export function isInternalDataPart(part: vscode.LanguageModelDataPart): boolean {
-  return part.mimeType === OPENCODE_USAGE_DATA_MIME || part.mimeType === COPILOT_USAGE_DATA_MIME;
+  return (
+    part.mimeType === OPENCODE_USAGE_DATA_MIME ||
+    part.mimeType === COPILOT_USAGE_DATA_MIME ||
+    part.mimeType === OPENCODE_REASONING_DATA_MIME
+  );
 }
 
 /**
--- a/src/commands/diagnostics.ts
+++ b/src/commands/diagnostics.ts
@@ -11,7 +11,15 @@ export async function showModelPickerDiagnostics(): Promise<void> {
   const sections: string[] = [];
 
   for (const vendor of vendors) {
-    const models = await vscode.lm.selectChatModels({ vendor });
+    let models: readonly vscode.LanguageModelChat[];
+    try {
+      models = await vscode.lm.selectChatModels({ vendor });
+    } catch (error) {
+      // One failing vendor (e.g. no Copilot models installed) must not abort
+      // the whole diagnostics report.
+      sections.push(`## vendor: ${vendor}`, "", `selection error: ${error instanceof Error ? error.message : String(error)}`, "");
+      continue;
+    }
     sections.push(`## vendor: ${vendor}`, "", `models: ${String(models.length)}`, "");
     for (const model of models) {
       const internalModel = model as unknown as { configurationSchema?: unknown; detail?: unknown };
--- a/src/commands/thinkingPicker.ts
+++ b/src/commands/thinkingPicker.ts
@@ -1,19 +1,20 @@
 import * as vscode from "vscode";
 import { CONFIG_SECTION } from "../config";
-import { getSettings } from "../provider/settings";
-import type { ThinkingSettings } from "../thinking";
+import { getSettings, THINKING_ALLOWED_VALUES } from "../provider/settings";
 
 /** Pick a model family then set its Thinking effort (writes config). */
 export async function showThinkingEffortPicker(): Promise<void> {
-  const families: { label: string; key: keyof ThinkingSettings; options: string[] }[] = [
-    { label: "DeepSeek (deepseek-v4-*)", key: "deepseek", options: ["off", "low", "medium", "high", "max"] },
-    { label: "GLM (glm-5, glm-5.1, glm-5.2)", key: "glm", options: ["off", "high", "max"] },
-    { label: "Kimi (kimi-k2.*)", key: "kimi", options: ["on", "off"] },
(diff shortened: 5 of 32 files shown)


Fahad090NP:
Hi @ltmoerdani 👋 — I know this is a big PR, so a few words up front to make it easy.

**1. It's 35 commits because they're deliberately one-fix-each and tiny.** No mega-diff hiding refactors — every commit is a single, self-contained bug fix with its own message (`fix(area): …`). You can review commit-by-commit and safely skip what you don't care about; nothing depends on anything else.

**2. Scope is strictly corrective, not creative.** I did *not* restructure anything. The largest single file change is the tool-schema cycle guard (~30 lines + tests); everything else is a few lines. No renames, no reorgs, no behavior I invented.

**3. Every commit passed the pre-commit gate, and the full `npm run lint` (eslint strict + tsc + prettier + markdownlint + shellcheck + all unit tests) is green on the merged head.** CI runs on this PR too.

**4. Honesty about confidence.** The description separates *verified* fixes from the ones that really want a quick live smoke test:
   - **Needs a live run before release**: the 5xx `Router.Unavailable` retry path, Google/Gemini tool-call streaming, vision-proxy empty-description path, and the DeepSeek reasoning-marker echo.
   - **Verified by reading + unit tests**: everything else (the two crash fixes have dedicated tests, plus regression tests for the day-bucket, schema, and utils changes).

**5. A few things I deliberately did NOT touch**, so you're not surprised they're absent: the `contextWindowHook` global monkey-patching (required for the feature — kept), the autocomplete Go-only key resolution (documented), and `AGENTS.md`/docs (separate PR #156).

**6. Three "bugs" I found turned out to be false positives — I rejected them and noted why in the PR thread** (fragmented tool-call names are intentional + tested; the token-estimate "double count" isn't one; the Qwen vision-proxy thinking path was already guarded). I'd rather say "I checked and it's fine" than churn code for nothing.

If any fix feels wrong or too aggressive, I'm happy to drop or rework that single commit — they're all independent, so carving any out is trivial. Happy to walk through any specific one.

Fahad090NP:
### What these fixes prevent in the real world

**1. Crashes that users would actually hit**
- Clicking the **"Models" tab in the Usage panel crashed** (missing `esc()` function) — the panel would just break/error.
- A **tool with a circular schema** (some agents/MCP tools have these) could **crash the whole extension host** with a stack overflow — not just the chat, *everything* in VS Code's extension host.

**2. Requests that fail for no good reason**
- During a gateway hiccup (502/503), the retry logic **silently didn't retry the Router.Unavailable case**, so users saw errors they could have waited through.
- If the user **cancelled during a retry**, they got a scary "502 gateway error" instead of a clean "cancelled".
- A **typo in a Thinking setting** (e.g. `deepseek: "banana"`) sent a bad `reasoning_effort` → **every request 400s** with no clear reason.
- A **temperature above 2** in settings → 400 on models that reject it.

**3. Wrong or silent usage/cost data (the worst kind)**
- **Afternoon usage counted on the wrong day** — the Today/Yesterday/Codebase numbers were subtly wrong.
- **Negative percentages** ("-5% used") when a manual usage target over-corrected.
- **Brand-new or unknown models tracked as $0** — looked free, so users' real spend was underreported and surprises came at billing.
- **Heavy CLI users saw an empty Usage panel** (the SQLite read silently failed past 64 MB).
- **Phantom tokens** in estimates from internal reasoning markers — wrong token/cost figures.
- **Deleting a profile didn't fully erase it** — stale meters could come back.
- With **multiple API keys, the "active" account silently jumped around** — you could be looking at the wrong account's usage.

**4. Broken tool calling / multi-turn reasoning**
- **Google/Gemini tool calls had empty IDs** → tool calls could get mixed up or lost in agent loops.
- Tool-call names could **duplicate** (`getWeathergetWeather`) if a gateway re-sent the name.
- Reasoning text could be **silently dropped** in some paths — breaking DeepSeek-style multi-turn conversations (the recurring 400).

**5. Annoying / confusing UX**
- **Ghost-text autocomplete sometimes never appeared** (a stale request killed the new one).
- Turning off **auto-enable Agents window never reverted the settings** the extension had flipped.
- Vision proxy **empty results stripped images** with a misleading "proxy unavailable" message.
- **One bad vendor killed the whole Diagnostics report**.

**6. Money-display mistakes**
- Costs showed as **`$-5.00`** (sign in the wrong place); HTML escaping missed single quotes.

— Every one of these is now fixed or explicitly verified-not-a-bug; the list maps 1:1 to the commits in this PR.

ltmoerdani:
Hi @Fahad090NP,

Reviewed this the slow way: checked out the branch, ran the full unit suite (324/324 green), typecheck clean, then spot-checked the riskier claims against `main` and the live gateway before trusting the writeup. That last step doesn't always survive contact with a 35-commit PR, so credit where it's due: everything I checked held up.

- `esc()` missing in the Models tab: confirmed. `dashboard.ts` on main calls it three times with no definition. Real crash.
- The 5xx `Router.Unavailable` retry: also confirmed dead. `isTransientServerError` needs the body for non-502/503/504 5xx, but `consumedErrorBody` was only ever set on the 400 path, so the branch could never fire. Nice find.
- Day bucketing `round` vs `floor`: agreed. With day-aligned buckets, an afternoon event lands in tomorrow.
- Model drift checks out against the live list: `minimax-m3` is there, `ring-2.6-1t` is gone.
- The active-profile override is the fix I'm happiest about. `ensureProfileSync` running unconditionally from `provideLanguageModelChatInformation` has bugged me for a while, and this is a clean way to stop it.

Two small things before merge, both doc-level:

1. The description lists the `getWeathergetWeather` name-duplication fix under Streaming, but there's no change for it in the diff and `toolCallAccumulator.ts` is untouched. You already noted in the thread you rejected it as a false positive, so just drop that line so nobody hunts for a commit that doesn't exist.
2. The comment in `dashboard.ts` cites `(issue #51)`, but #51 is the model-list fetch notification spam. The relevant one is #63, the usage panel with two opencode profiles. Mind pointing the comment there? Happy to fix it myself on merge if you'd rather not rebase.

One judgment call I want to flag rather than ask you to change: `UNKNOWN_GO_MODEL_PRICE` (0.5 in / 2.0 out) plus the "family estimate" entries means unknown models now over-report instead of reading as free. $0 is the worse failure, I agree. But an over-report has its own cost: someone watching their cheap daily driver get priced like a premium model may stop trusting the tracker. Product decision more than a bug fix, and worth getting real prices into the snapshot once they land.

Plan: merge as a merge commit per repo convention (never squash, I want all 35 commits in the history), then live smoke test the streaming paths (5xx retry, Google tool calls, vision placeholder) before the next release. No code changes requested.

Thanks for this. The honesty in the checklist and the rejected false-positives writeup made reviewing a 35-commit PR a lot easier than it usually is.

Fahad090NP:
Both review notes are addressed:

1. Dropped the `getWeathergetWeather` line from the description — the false-positive note is gone so nobody hunts for a commit that doesn't exist.
2. Fixed the `dashboard.ts` comment: it now cites `(issue #63)` (usage panel with two opencode profiles) instead of #51 (model-list fetch notification spam). Committed as `docs(usage): point active-profile override comment at the right issue (#63)` on the branch.

The 5xx retry / Google tool-call / vision placeholder live smoke test plan still stands for after merge. Thanks for the thorough review.

ltmoerdani:
Hi @Fahad090NP,

Both confirmed on my end. Pulled the branch again, checked the new commit, and the `dashboard.ts` comment now points at #63. Grep on the PR description also comes back clean, so the `getWeathergetWeather` line is gone. Thanks for picking both up that fast.

Nothing else from me. The plan stands: merge as a merge commit (never squash, I want all 36 commits in the history), then live smoke test the streaming paths (5xx retry, Google tool calls, vision placeholder) before the next release.

Nice work on this one. The one-fix-per-commit structure and the honest checklist made it a genuinely easy PR to review.

PR #160, opened 2026-08-16
feat: add OpenAI thinking effort setting

buhagee:
## 📝 What does this change?

Adds the `opencodego.thinking.openai` setting for OpenAI GPT models, including support for `off`, `low`, `medium`, `high`, and `xhigh` reasoning effort. The setting maps to the Responses API `reasoning.effort` field.

## 🧪 How did you test it?

Validated the package configuration and ran the full test suite:

- `npm run compile`
- `npm run lint`
- `npm test`
- `npm run package`
- `npm run test-retry` — 7/7 passed

Live testing with an OpenAI model was not performed.

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works

### DIFF
--- a/package.json
+++ b/package.json
@@ -396,6 +396,18 @@
           "default": "off",
           "markdownDescription": "Thinking mode for MiniMax models (`minimax-m2.5`, `minimax-m3`, etc.). Maps to `thinking: { type: 'disabled'|'adaptive'|'enabled' }`. The OpenCode gateway only supports on/off for this family (`reasoning_effort` is silently ignored)."
         },
+        "opencodego.thinking.openai": {
+          "type": "string",
+          "enum": [
+            "off",
+            "low",
+            "medium",
+            "high",
+            "xhigh"
+          ],
+          "default": "off",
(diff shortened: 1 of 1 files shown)


ltmoerdani:
Nice one @buhagee, thanks for this! This closes a gap we missed on our side: the OpenAI thinking strategy, the config key, and the defaults all landed earlier with the per-provider thinking refactor, but the `package.json` declaration never made it in, so the setting stayed invisible in the Settings UI.

I checked the details as well. The enum values match `OPENAI_EFFORTS` in `src/thinking/openai.ts` and the default matches `THINKING_DEFAULTS`, so it's declaration-only with no behavior change. The payload shape (`reasoning: { effort }`) is already covered by unit tests, so skipping the live model test is fine here.

One small note, nothing blocking: the settings table in `docs/features/02-20260517-per-model-thinking-controls.md` doesn't have a row for this setting yet. I'll add that myself after merging, no action needed on your end.

Merging this with a merge commit.

PR #161, opened 2026-08-17
feat: restore API key command and route usage

buhagee:
## 📝 What does this change?

Restores optional command-palette API-key entry for OpenCode Go and OpenCode Zen. Keys are stored securely in VS Code SecretStorage, native VS Code BYOK remains supported, and models refresh after saving.

Also routes model, test-connection, and Go usage requests through the configured `apiBaseUrl`, including safe usage-sync diagnostics without exposing credentials.

## Why is this change needed?

The extension supports configurable OpenCode-compatible API gateways, but not all traffic was consistently routed through the configured base URL. Chat/model requests used `opencodego.apiBaseUrl`, while Go usage synchronization still used the hardcoded OpenCode endpoint. As a result, users running a compatible gateway could receive valid model responses while their usage dashboard incorrectly showed zero.

This change makes the configured API base URL authoritative for model discovery, connection testing, chat requests, and usage synchronization.

It also restores an optional command-palette API-key entry path. Native VS Code BYOK remains supported and is still the preferred flow, but the fallback is useful when the BYOK panel is unavailable or when users are testing a custom OpenCode-compatible gateway. Keys remain in VS Code SecretStorage, and models refresh immediately after a key is saved.

## 🧪 How did you test it?

- Tested OpenCode Go with `gpt-5.6-luna` through the configured custom gateway.
- Confirmed requests use the custom API base URL.
- Confirmed Go usage no longer reports zero when the gateway provides the compatible `/usage` endpoint.
- Confirmed API-key storage and model refresh behavior.
- `npm run compile` passed.
- `npm test` passed.
- `npm run test-retry` passed: 7/7 tests.
- `npm run package` passed and produced the VSIX.
- Staged ESLint and Prettier checks passed.

`npm run lint` was also run, but the repository-wide check reports pre-existing line-ending and formatting issues across unrelated files. ESLint passed for the changed files.

## ✅ Checklist

- [ ] `npm run lint` passes — existing repository-wide Editorconfig/Prettier issues remain
- [x] `npm test` passes
- [x] I tested it works


ltmoerdani:
Hey @buhagee, thanks for this. I checked the routing change against the code and it holds up. Usage sync now reads `opencodego.apiBaseUrl` the same way the provider does, and `setApiKey` writes to the per-vendor secret, so the Go/Zen collision we fixed in `379d86a` stays fixed.

One thing I want to settle before merge. This restores the command-palette API-key entry we deliberately pulled a few days ago, when we went BYOK-only in `379d86a`. I'm not against having it back. I'd just like to know what's driving it, so the scope lands in the right place.

Is it the environment where the native Manage Models panel isn't available? Or a specific flow where BYOK doesn't work for you? The answer decides whether this stays a documented fallback or gets wired deeper, say into the multi-profile flow from issue #63.

ltmoerdani:
Hey @buhagee, this new commit answers the question from my last comment, so thanks for that. Multi-profile is the right home for this, and registering the profile inside `setApiKey` matches what I had in mind. I pulled the branch, compile is clean on my side.

One gap though. `setActiveProfile()` reads a per-profile secret:

```ts
const apiKey = await extensionContext().secrets.get(profileSecretKey(fingerprint));
```

but nothing in the branch ever writes `opencodego.apiKey.profile.*`. I grepped the tree, and apart from the definition that `get` is the only call site. It always resolves to `undefined`, so the `if (apiKey)` branch never runs, and after a window restart a profile switch still falls back to the base secret, which may belong to another account. The bug the commit message describes is still there.

The clean writer spot is `ensureProfileSync()` in `dashboard.ts`. Every path that learns a key now flows through it: `setApiKey`, model discovery, and request recording via `ensureProfileForApiKey`. One fire-and-forget line covers all three:

```ts
void extensionContext().secrets.store(profileSecretKey(fp), apiKey);
```

A small test would lock it in, something asserting `setActiveProfile` picks up the stored key instead of the base secret.

One question so I understand your setup: when more than 1 key wasn't registering, what did that look like on your end? Profile missing from the picker, or wrong profile staying active after a restart? Want to make sure the writer above covers your case before you spend time on it.

PR #163, opened 2026-08-18
Fix infinite looping for MiMo

Barragek0:
## 📝 What does this change?

Adds `repetition_penalty: 1.2` to MiMo payloads to fix infinite-thinking-loop observed in MiMo v2.5 / v2.5 Pro.

https://github.com/XiaomiMiMo/MiMo-Code/issues/914 - Important excerpt from the `Root Cause Analysis` section: NVIDIA community notes that repetition_penalty=1.2 mitigates MiMo "Thought Loop" failure mode

Contrary to most information I found on this issue, it does seem to happen when reasoning is off in MiMo too, it just seems to be less frequent, so I added it for both reasoning on and off.

## 🧪 How did you test it?

Tested with **MiMo v2.5 / v2.5 Pro** on **High** and **Off** reasoning effort, no longer loops endlessly.

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes (324/324)
- [x] I tested it works
- [x] I updated docs/CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,7 +30,7 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ### Fixed
 
-- **`[Thinking]` Sampling-level `repetition_penalty` for MiMo curbs infinite thinking loops (#36, https://github.com/XiaomiMiMo/MiMo-Code/issues/914).** A `repetition_penalty: 1.2` is now applied unconditionally in every MiMo payload (regardless of reasoning effort) to suppress the degenerate token-repeat pattern that triggered the Go gateway's infinite-loop detector upstream. This is a third mitigation layer alongside the `budget_tokens` cap and suffix-repetition stream detection shipped earlier. Applies with reasoning off (where the loop fires less often) and on; does not trip `bodyRequestsThinking()` so the gateway workaround path is unaffected.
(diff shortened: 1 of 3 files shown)


ltmoerdani:
Nice one @Barragek0, this is a clean fix and the upstream research checks out.

Verified on my side: checked out the branch, tests 324/324, compile clean. Also confirmed `repetition_penalty` doesn't trip `bodyRequestsThinking()`, so the Go gateway #37635 workaround path is unaffected.

Context for anyone reading later: this is a third mitigation on top of what we shipped for #36 (budget_tokens cap + suffix-repetition detection in the stream extractor). A sampling-level penalty makes sense, and applying it with reasoning off matches the upstream analysis, the loop just fires less often there.

Two things:

1. Checklist says "I updated docs/CHANGELOG" but the diff doesn't touch `CHANGELOG.md`. Can you add an entry? Happy to handle it on merge if that's easier, just let me know.

2. Small nit, not a blocker: consider pulling `1.2` into a named constant next to `MIMO_BUDGET_MAP` (e.g. `MIMO_REPETITION_PENALTY`) so it reads as a tuning knob instead of a magic number. Fine to keep inline if you prefer.

Once the CHANGELOG entry is sorted I'll merge with a merge commit so your three commits stay intact.

PR #164, opened 2026-08-18
Add force think tag stripping to fix blank code boxes appearing while subagents are running

Barragek0:
## 📝 What does this change?

Force think-tag stripping when tools are present in the request (subagent/tool-call invocations). This prevents <think> tags in content from rendering as blank code blocks in the chat UI when using a subagent, flooding the chat with blank code boxes.

The trigger is tool-call detection (options.tools.length > 0), not model-specific. The ThinkTagFilter is a no-op for models that don't emit think tags, so this is safe for all models.

## 🧪 How did you test it?

Tested with **MiMo v2.5**

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works

### DIFF
--- a/src/core/transport.ts
+++ b/src/core/transport.ts
@@ -42,6 +42,12 @@ export interface StreamRequestOptions {
    * - "always" — strip for every model
    */
   stripThinkTags?: "never" | "auto" | "always";
+  /**
+   * Force think-tag stripping regardless of stripThinkTags mode.
+   * Set to true when the request is a subagent/tool-call invocation
+   * to prevent <think> tags in content from rendering as blank code blocks.
+   */
+  forceStripThinkTags?: boolean;
 }
 
 export interface TransportRequestSummary {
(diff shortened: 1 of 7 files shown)


ltmoerdani:
Thanks for this one @Barragek0, clean approach. Reusing the existing `stripThinkTags` plumbing and forcing the mode when tools are present beats a model-specific hack, and the no-op passthrough for models that don't emit think tags makes it safe across the board.

I pulled the branch and verified beyond lint: the full `npm run lint` gate passes (all 7 checks), and I ran a direct logic check on `createThinkTagFilter`. `forceOverride` activates the filter even on `never`, non-forced requests keep their old behavior, and the visible/thinking split comes through intact. Also like that the stripped thinking still surfaces via `LanguageModelThinkingPart`, so nothing gets dropped.

One small follow-up I'll handle myself after merging: a tiny unit test for the force path, since `thinkTags.ts` has no coverage yet. Not a blocker.

Merging now. Thanks again!

Issue #165, opened 2026-08-19
[BUG] Muse Spark 1.2 model not working


harshmann10:
# 🐛 Muse Spark 1.2 — API 400 Error on Every Request

## 🐛 What happened?

Muse Spark 1.2 (`muse-spark-1.2-contributor`) fails on every request with HTTP 400. The API returns an empty assistant response with `finish_reason: null`, causing the extension to throw `OpenCodeRequestError`. The model appears to reject the request payload (990KB) without providing a meaningful error message.

## 🔄 Steps to reproduce

1. Set provider to **OpenCode Go**
2. Select model **muse-spark-1.2-contributor**
3. Send any prompt (even a simple one like "hello")
4. Request fails immediately with: `Sorry, your request failed. Please try again.`

## ✅ What did you expect?

The model should respond with a completion, or return a descriptive error message explaining why the request was rejected (e.g., payload too large, invalid parameters, model unavailable).

## 🖥️ Environment

- **Extension version:** 0.6.0
- **VS Code version:** (please fill in — Help → About)
- **Provider:** OpenCode Go
- **Model:** muse-spark-1.2-contributor

### Error 1 (user-facing):
```
Sorry, your request failed. Please try again.

Client Request Id: f84dae2f-0b63-49cd-a545-60aac4b4b2a0

Reason: OpenCode Go API request failed (400) model=muse-spark-1.2-contributor payloadBytes=990778: {"id":"chatcmpl_oikufvx92vl","object":"chat.completion","created":1787146321,"model":"muse-spark-1.2-contributor","choices":[{"index":0,"message":{"role":"assistant"},"finish_reason":null}]}
```

### Error 2 (console):
```
OpenCode Go API request failed (HTTP 400) for muse-spark-1.2-contributor: {"id":"chatcmpl_oikufvx92vl","object":"chat.completion","created":1787146321,"model":"muse-spark-1.2-contributor","choices":[{"index":0,"message":{"role":"assistant"},"finish_reason":null}]}
```

### Stack trace:
```
at buildOpenCodeRequestError (c:\Users\Harsh\.vscode\extensions\ltmoerdani.opencode-copilot-chat-0.6.0\out\errors.js:39:12)
at streamOpenCodeResponse (c:\Users\Harsh\.vscode\extensions\ltmoerdani.opencode-copilot-chat-0.6.0\out\streaming.js:360:73)
at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
[...]

PR #167, opened 2026-08-20
fix: trim conversation history to fit context window (prevents 'No response came')

Fahad090NP:
## Summary

Fixes two related "No response came" failure modes with a single root-cause fix:

1. **Multiple VS Code windows** — chatting in separate windows intermittently fails with no response.
2. **Repeated failures fixed by Compact Conversation** — after many turns the request fails every time; running *Compact Conversation* and sending one message works again.

### Root cause

`provideLanguageModelChatResponse` sent the **entire** conversation history as the request payload. The only history guard was image trimming (`trimOldImagesFromHistoryInPlace`) — there was **no text-history truncation**. So a long conversation grows past the model's context window and the upstream either:
- rejects the oversized request with HTTP 400 (empty assistant message, `finish_reason: null`), or
- returns an empty stream → VS Code surfaces **"No response came"**.

Compact Conversation shrinks the history, which is exactly why it "fixes" the symptom. (This also explains the reported Muse Spark ~990 KB payload — it is the uncompacted history, not a model-specific quirk.)

### Changes
- **`src/provider/historyTrim.ts`** (new, pure) — `trimOldMessagesToFitContext` drops the oldest messages until the estimated prompt tokens fit `contextWindow − maxOutputTokens − safetyMargin`. It **preserves** the first (system/anchor) and last (current prompt) messages, and **never splits a tool-call group** (stops before the first assistant message carrying `tool_calls`), so no tool reference is orphaned.
- **`src/provider/OpenCodeProvider.ts`** — calls it right after the image-history trim, before the prompt-token estimate, with a `[history-trim]` diagnostic log.
- **`src/config.ts`** — `HISTORY_TRIM_SAFETY_MARGIN_TOKENS = 2048`.
- **Tests** — `src/test/messages.test.ts` (no-op when fitting, drops oldest keeping anchor + last, never splits a tool group, no-op when only anchor + last remain).
- **Docs** — CHANGELOG Fixed entry + `docs/issues/68-20260820-history-trim-context-overflow.md`.

### Verification
- `npm test` — 337/337 pass
- `npm run lint` — all 7 gates pass (Editorconfig, ESLint, Markdown, Prettier, Shell, TypeScript, Tests)

PEACE BE UPON YOU

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 - **DeepSeek / Mimo thinking content no longer leaks into the chat transcript.** `treatReasoningAsContent` was mis-detecting native-reasoning families as "no reasoning in body" and echoing their `reasoning_content` as plain chat text. The decision now comes from the provider strategy (always `false` for DeepSeek and Mimo), so chain-of-thought stays in the thinking panel.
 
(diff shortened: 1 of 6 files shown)


Fahad090NP:
Superseded by #169 — that PR is branched from this one and adds the hardening (70% context ratio, 512 KB byte cap, O(n) trim, tool-group dropping) that resolves the 503 / 10-min timeout / empty-response / session-slowness symptoms and closes #165.

PR #168, opened 2026-08-20
Add Muse Spark 1.2 models

### LINKED ISSUE #165: [BUG] Muse Spark 1.2 model not working

Barragek0:
## 📝 What does this change?

Adds full support for Muse Spark 1.2 models (muse-spark-1.2-contributor on Go, muse-spark-1.2-contributor-free on Zen). Closes #165.

## 🧪 How did you test it?

Tested with `muse-spark-1.2-contributor` (Go) and `muse-spark-1.2-contributor-free` (Zen) - verified chat responses, tool calling, and thinking effort levels (off/high). All 341 unit tests pass, lint is green, VSIX packages cleanly.

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ### Added
 
+- **`[Models]` Muse Spark 1.2 support.** Full support for `muse-spark-1.2-contributor` (Go) and `muse-spark-1.2-contributor-free` (Zen). Muse models are routed through the Responses API with `truncation: disabled` and tool names truncated to 64 characters (both gateway constraints). A new `MuseThinking` strategy provides configurable thinking effort (off/low/medium/high/xhigh) via `reasoning: { effort }`, and the thinking effort picker now includes a Muse Spark entry. Fallback metadata provides 1M context / 131K output limits when the live models.dev fetch is unavailable.
+
 - **`[Providers]` Configurable API base URL.** The extension no longer hardcodes the `opencode.ai` endpoints — it can now point at any compatible gateway via `opencodego.apiBaseUrl` (default `https://opencode.ai/zen/go/v1`) and `opencodezen.apiBaseUrl` (default `https://opencode.ai/zen/v1`). The extension derives the `/chat/completions`, `/messages`, `/responses`, `/models`, and (Go only) `/usage` routes from the configured base and applies them to every provider (normal chat, Agents window variants, inline completions, usage sync). Custom URLs are normalized and validated — non-`http(s)`, embedded credentials, query strings, and hashes are rejected and fall back to the default. Reload the window after changing the setting. Unit tests added for URL normalization and route construction.
 
 ---
--- a/package.json
+++ b/package.json
@@ -440,6 +440,18 @@
           ],
           "default": "auto",
           "markdownDescription": "Optional `thinking_budget` for Qwen models when thinking is `on` or `auto`. `auto` omits the field. Ignored when Qwen thinking is `off`."
+        },
+        "opencodego.thinking.muse": {
+          "type": "string",
+          "enum": [
+            "off",
+            "low",
+            "medium",
+            "high",
+            "xhigh"
+          ],
+          "default": "off",
+          "markdownDescription": "Thinking mode for Muse Spark models (`muse-spark-1.2-contributor`, `muse-spark-1.2-contributor-free`). Maps to the Responses API `reasoning.effort` field. Higher effort increases reasoning depth and may increase latency and usage."
         }
       }
     },
--- a/src/commands/thinkingPicker.ts
+++ b/src/commands/thinkingPicker.ts
@@ -15,6 +15,7 @@ export async function showThinkingEffortPicker(): Promise<void> {
     { label: "OpenAI GPT (gpt-*)", key: "openai", options: [...THINKING_ALLOWED_VALUES.openai] },
     { label: "Qwen (qwen3.*)", key: "qwen", options: [...THINKING_ALLOWED_VALUES.qwen] },
     { label: "Qwen Thinking Budget", key: "qwenBudget", options: [...THINKING_ALLOWED_VALUES.qwenBudget] },
+    { label: "Muse Spark (muse-spark-*)", key: "muse", options: [...THINKING_ALLOWED_VALUES.muse] },
   ];
   const settings = getSettings().thinking;
   const family = await vscode.window.showQuickPick(
--- a/src/config.ts
+++ b/src/config.ts
@@ -60,6 +60,7 @@ export const SETTING_THINKING_OPENAI = "thinking.openai";
 export const SETTING_THINKING_QWEN = "thinking.qwen";
 export const SETTING_THINKING_QWEN_BUDGET = "thinking.qwenBudget";
 export const SETTING_THINKING_MIMO = "thinking.mimo";
+export const SETTING_THINKING_MUSE = "thinking.muse";
 
 // ─── Inline completions (issue #49) ─────────────────────────────────────────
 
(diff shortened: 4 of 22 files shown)


Barragek0:
Thinking blocks are not appearing, and I'm not sure why. The code all looks correct, so I think it must be an upstream issue with the OpenCode gateway, which isn't relaying `reasoning_content` in the stream, so the extension has nothing to capture, unless I'm just misunderstanding the codebase, which is probably more likely.

ltmoerdani:
@Barragek0 This PR is clean. Routing through the registry, the new thinking strategy, fallback metadata, tests per layer: all of it follows the existing patterns. I compiled and ran the tests locally on your branch too, 341/341 pass.

One thing I spotted before merging: **the truncated tool name is never mapped back**.

`truncateToolName` runs at request build time, so the model only ever knows the truncated name. When the model returns a tool call, the extractor in `src/transports/extractors.ts` emits a `LanguageModelToolCallPart` with that truncated name, exactly as it comes back. VS Code resolves the tool by name. For tools over 64 chars (MCP tools hit that often, especially with a server prefix), no match against the original registered tool, so the call fails.

We already solved this shape of problem once. `src/request/google.ts` keeps a truncated → original name map (`toolNamesById`) for the round trip, and something similar fits here: store a `truncated → original` map at request build time, then reverse-lookup in the responses extractor before emitting the tool call part.

Two minor notes, not blockers:
- `/^muse-/i.test(modelId)` is duplicated in two places in `openai.ts` even though the registry already has `family: "muse"`. Could derive it from the registry to keep a single source of truth.
- `truncation: "disabled"` means requests over 1M context will hard-fail instead of being truncated upstream. Reasonable trade-off, just worth a line in the docs.

About the thinking blocks not showing: I checked too, and I agree with your read. The request payload is correct, most likely the gateway just isn't relaying `reasoning_content` in the stream. Not an extension-side issue. I'll note it in #165.

Just that one fix and the rest is good to merge 👍

Barragek0:
All three points addressed in the latest commits:

**Truncated tool name round-trip** - `src/request/openai.ts` now exports `buildResponsesToolNameMap()`, which builds a `truncated → original` map at request build time. The map is threaded through `StreamRequestOptions.toolNameMap` → `streamResponsesApi()` → `OpenAiResponseExtractor`, which reverse-lookups before emitting `LanguageModelToolCallPart`. Mirrors the `toolNamesById` pattern in `src/request/google.ts`. Covered by new tests in `src/test/openai-request.test.ts` (`buildResponsesToolNameMap` suite) and `src/test/extractors.test.ts` (extractor round-trip).

**`/^muse-/i` duplication** - Replaced both inline regex checks with `isMuseFamily()`, which derives the result from the registry (`lookupModelRegistryEntry(modelId).family === "muse"`). Single source of truth, one place to touch.

**`truncation: "disabled"` docs** - Added a comment in `buildResponsesRequestBody()` explaining it's a gateway constraint: requests exceeding the 1M context window will hard-fail (HTTP 400) instead of being silently truncated upstream.

Build, lint (7/7), tests (350/350), and VSIX packaging all pass. 🟢

Barragek0:
One issue to note, potentially blocking merge, the Context Window in Copilot does not correctly update when the new models are used. The Context Window progress circle element remains at 0%. When reaching the 1M context window limit, a new chat needs to be opened as the conversation cannot be compacted. Will look into fixing this.

Edit: Two separate issues, which I believe are a problem with the model and not the branch, spacing is sometimes incorrect in responses, there are often missing spaces and the model repeatedly re-sends the same response in between tool calls, example:

```
Found the error zone — now tracing the unclosed string to its source.

Created check_parse.ps1

Found the error zone — now pinpointing the exact line.

Powershell tool call

Found the error zone — now pinpointing the exact line.

Powershell tool call

Found the error zone — tracing the unclosed string to its source.
```

I've looked into both and again, everything looks correct on the extensions' end, leading me to believe they are problems with the model itself, and not the implementation.

ltmoerdani:
Agreed, let's keep this PR scoped to what it sets out to do: getting muse-spark-1.2-contributor (Go) and muse-spark-1.2-contributor-free (Zen) working, which is exactly what #165 asked for. The Responses API routing fixes the 400, and I verified the full path locally: metadata matches models.dev on both providers, tool name round-trip is solid, 350/350 tests pass.

On the two things you flagged, I dug into both and neither traces back to this branch:

- **Context window not updating**: the advertised limits path is correct end to end. Fallback metadata, live models.dev (both providers now list the muse models at 1M/131K), and the advertised calculation all agree, and the model version string already bakes in the limits so VS Code should invalidate stale metadata. This looks like VS Code-side caching. If you can share your VS Code version plus what number the picker shows vs what you expected (and whether a window reload changes it), I'll track it as a separate issue.
- **Spacing / repeated responses**: the repetition pattern looks like upstream model behavior, we've seen a similar echo issue with another model family before (reasoning content being relayed as regular text). Not something this PR can or should fix.

So from my side this is good to merge as-is. I'll file the context window thing separately if the repro details come in. Thanks for the quick turnaround on the review points 🙌

Barragek0:
Found the problem after debugging the responses, the gateway is sending usage `input_tokens` in `response.completed`, but we aren't reading it. Will open a new pull request.

PR #166, opened 2026-08-20
fix: GLM 5.3 always-thinking (force-on) — fixes #162

Fahad090NP:
## Summary

Fixes #162 — GLM 5.3 (and later) rejects `thinking: { type: "disabled" }` with HTTP 400:
> "This model always engages in thinking and cannot be disabled; please use low, high, or max"

### Changes
- **`src/thinking/glm.ts`** — `GlmThinking` now detects GLM 5.3+ by version (`/^glm-5[.\-](?:[3-9]|\d{2,})/i`) and forces thinking on (default `high`). The model picker exposes only `high`/`max` and hides `off`, mirroring the existing Kimi K2.7-code force-on behavior. `normalize()` upgrades any stale `off` override to `high` so cached settings can't trigger the 400.
- **`src/retry.ts`** — added a defensive recovery pattern: when the upstream reports it "cannot be disabled" / "always engages in thinking", the `thinking` field is stripped from the body so a retry succeeds instead of looping.
- **Tests** — `src/test/thinking.test.ts` (version detection, forced-on resolution, payload shape, picker schema) and `src/test/retry.test.ts` (cannot-be-disabled recovery).

### Verification
- `npm test` — 333/333 pass
- `npm run lint` — all 7 gates pass (Editorconfig, ESLint, Markdown, Prettier, Shell, TypeScript, Tests)

PEACE BE UPON YOU

PR #169, opened 2026-08-20
fix: bound history to 70% context + 512KB cap (503/timeout/empty-response/slowness)

Fahad090NP:
fix: bound history to 70% context + 512KB cap (503/timeout/empty-response/slowness)
===
## Summary

Fixes the full symptom set (503 / request timeout / empty-response / session slowness) — all rooted in an **unbounded conversation-history payload**:

- **`503 Endpoint is unavailable`** (`payloadBytes` ≈ 783 KB)
- **`request timed out after 10m 0s`** (upstream hangs on a huge payload)
- **"No response came" / "Sorry, no response was returned"** (empty stream)
- **Session slowness** specific to the VS Code extension (OpenCode / other clients fine)

The reporter observed failures begin at **~70% context full** and that *Compact Conversation* fixes them — confirming the payload (not the model) is the cause.

## Root cause

`provideLanguageModelChatResponse` sent the entire history every turn. The only guard was image trimming. Two gaps remained:

1. The token budget was `context − output − margin`, which can still sit **above** the ~70% failure threshold, so trimming to "the window" alone didn't always drop enough.
2. The trim re-estimated the **whole** payload on every candidate drop (**O(n²)**) — the per-request session slowness.

## Changes

- **`src/config.ts`** — `HISTORY_TRIM_TARGET_RATIO = 0.7` (stay safely below the observed failure threshold) and `MAX_REQUEST_PAYLOAD_BYTES = 512 KB` (hard byte backstop for inaccurate token estimates).
- **`src/provider/historyTrim.ts`** — `trimOldMessagesToFitContext` now bounds on **both** a token budget and the byte ceiling, runs in **O(n)** (estimate once, subtract per dropped unit), drops complete **tool-call groups as one unit** (never orphaning a reference), preserves anchor + current prompt, and returns `{ removed, finalTokens, finalBytes }`.
- **`src/provider/OpenCodeProvider.ts`** — trims to the 70% ratio + byte cap and reuses the returned estimate (no second re-estimation).
- **`src/test/messages.test.ts`** — extended: byte-cap enforcement, tool-group dropping, returned estimates, anchor/prompt preservation.
- **Docs** — CHANGELOG Fixed entry + `docs/issues/68-20260820-history-trim-context-overflow.md`.

## Verification

- `npm test` (all pass) and `npm run lint` (all 7 gates) green.
- Staged-lint pre-commit hook passes on every commit.

PEACE BE UPON YOU

---

## Also includes: chat-request fetch() resilience (issue #78)

This PR now also contains the fix for transient `fetch()` failures on the chat request path (the `Reason: fetch failed: TypeError: fetch failed` error). The engine now retries transient network throws via `fetchWithTransientRetry`, mirroring the model-list fetch retry from issue #78. See `docs/issues/69-20260820-chat-request-fetch-resilience.md`.


ltmoerdani:
I checked out your branch @Fahad090NP and ran the full lint stack locally: all 7 gates pass, and the O(n) subtractive trimming with tool-group atomicity is well built. This is getting merged.

The token-budget trimming solves a real one for us: #109 hit "maximum context length exceeded" exactly because the full history went out with no text trim. The fetch retry is a clean win too, reusing the #78 classifier the right way.

Two notes for follow-up, neither blocking:

1. The 512KB byte cap doesn't match what we've measured. #104 was closed withdrawn after we confirmed the gateway accepts bodies up to ~1.5MB (see #44), so from our side there's no hard limit at that size. I'm leaving the cap in as a backstop, but I'll open an issue for the real edge case: with `MAX_HISTORY_IMAGES_KEPT = 2` and images up to 5MB base64, a vision payload can sit above 512KB even after token trimming, and the trimmer would drop the entire middle history for nothing. Likely fix is excluding image data from the byte accounting.

2. The issue references need cleanup: "Closes #165" points at the Muse Spark report (already handled by #168) and "Refs #69" points at the usage-window bug — the fetch-resilience one is #78, which your commit message already cites correctly. The two docs (`docs/issues/68-...` and `69-...`) have the same mixup. A small commit fixing those would keep the history searchable.

Merging with a merge commit as always. Thanks for this one.

Fahad090NP:
References cleaned up in 98f6a7c (+ amended trailer on the fetch-resilience commit):

- `Refs #69.` → `Refs #78.` — #69 is the usage-window PR, you're right; the fetch-resilience work refs #78.
- Both docs now link their actual related issues: `68-…history-trim…` → #109 (same unbounded-history root cause), `69-…fetch-resilience…` → #78 (same classifier, extended to the chat path). History is searchable by issue number again.
- The stray "Closes #165" is already gone from the PR description.

On your 512KB note — agreed, and thanks for leaving the cap as a backstop. Excluding image data from the byte accounting (or raising the cap for vision payloads) sounds like the right shape; happy to pick that up when you open the issue.

ltmoerdani:
hey, quick note before this lands: main had moved under your branch. PR #166 (GLM 5.3 force-on thinking) added an entry to the same CHANGELOG Fixed section your two `[Chat]` entries go in, so GitHub was flagging this PR as conflicting.

I resolved it on my side. Merged your branch into main locally, kept both sides of the one `CHANGELOG.md` conflict (main's GLM entry first, then your two), and ran the full lint stack on the result: all 7 gates pass, tests included. Everything else auto-merged clean. Merging now with a merge commit, so all 12 of your commits land intact.

Thanks for the quick turnaround on the references, and for this one generally. The token-trim against #109 is the part I'm most glad to have.

PR #170, opened 2026-08-20
fix(streaming): detect truncated streams and complete promptly on [DONE]

Fahad090NP:
## Summary
The shared streaming engine (`src/transports/engine.ts`) only ended a response stream on **connection close** or **user cancellation**, ignoring OpenCode's `data: [DONE]` terminator (confirmed in `tmp/opencode-dev/.../server/transport/ws.ts`, which enqueues `data: [DONE]\n\n` then closes). This produced two silent failure modes that looked like "the model stopped working":

1. **Hang-then-cancel** — if the gateway kept the connection alive after `[DONE]`, the read loop never saw `done` and sat until the 2-minute idle timeout (user cancelled manually, no error shown).
2. **Silent truncation** — if the connection dropped *before* `[DONE]` (proxy reset, upstream crash, truncated payload, VPN/firewall cut), `reader.read()` returned `done` with no `[DONE]` and no `finish_reason`; the engine treated it as a successful empty response.

## Fix
- `src/transports/sse.ts` — `parseServerSentEvent` now takes an `onDone` callback fired on `data: [DONE]` (and `[DONE]` is no longer passed to the extractor).
- `src/transports/engine.ts` — tracks `streamFlags.sawDone` and **breaks the read loop as soon as `[DONE]` arrives** (prompt completion instead of waiting on a lingering socket); after the loop, `isStreamTruncated` throws a clear `OpenCodeRequestError` when the stream ended with neither `[DONE]` nor a captured `finish_reason` after content was already received, instead of silently succeeding.
- `src/transports/sse.ts` — `isStreamTruncated` is a small pure helper (exported, unit-tested).
- `src/test/sse.test.ts` — unit tests for `[DONE]` handling and `isStreamTruncated`.

## Verification
- `npm run lint` (editorconfig, eslint, markdown, prettier, shellcheck, typecheck, unit tests) — green.
- `src/test/sse.test.ts` — new tests pass.

Builds on the resilience work in #169 (history-trim + transient fetch-retry).


ltmoerdani:
@Fahad090NP the diagnosis matches what I see in `engine.ts` — the loop really only exits on connection close or cancellation, so both failure modes track. The truncation guard conditions (`extractedPartCount > 0 && totalBytes > 0`) also look right, no false positives on empty streams.

A few things before I merge:

**1. Do the Google and Anthropic routes actually send `[DONE]`?**

This is the main one. The OpenCode chat-completions gateway terminates with `data: [DONE]`, but we route 4 transports through this engine:

- Google `:streamGenerateContent?alt=sse` is native Google SSE — I don't believe it uses the OpenAI `[DONE]` convention
- Anthropic `/messages` ends with `message_stop`, not `[DONE]`

For those two, truncation detection falls back entirely on `finishReason`. Problem: `normalizeGoogleFinishReason()` returns `null` for values like `OTHER`, and `updateRequestUsageSummary()` only records `finishReason` when it's a string. So a healthy Gemini response that ends with a finishReason we normalize to `null` would hit `isStreamTruncated() === true` and throw on a response that was actually fine.

Can you either (a) show a live SSE capture for the Google + Anthropic routes proving `[DONE]` is sent / `finishReason` is always captured, or (b) gate the truncation check to transports we've confirmed send `[DONE]`?

**2. Socket cleanup on break**

When we break the loop on `sawDone` while the socket is still alive, we rely on the engine's `finally` aborting the controller. I believe it does, but worth confirming there's no path where the connection lingers after the break.

**3. Flush paths get skipped on throw**

In `chatCompletions.ts` / `responses.ts` / `google.ts`, `flushRemainingToolCalls()` and `flushReasoningFallback()` run after `await streamOpenCodeResponse(...)` with no try/finally. When the truncation error throws, pending tool calls and accumulated reasoning are dropped. Not a blocker since the error message tells the user to retry, but moving the flushes into a `finally` would be cleaner.

Also heads up on history: the Go gateway has sent `finish_reason: null` on final events before (gpt-5.6-luna, #93/#41). With this change those streams will now surface an explicit error instead of silently completing, which is the intent, but expect a few more error reports if the gateway glitches.

Points 2 and 3 can be follow-ups. Point 1 I'd like settled before merge.

Fahad090NP:
All three points settled:

**1. [DONE] gating (blocking) — done in 5983a3b, option (b).** Added a per-transport `usesDoneSentinel` flag: `true` only for chat-completions and Responses; `false` for Google (`:streamGenerateContent?alt=sse`) and Anthropic (`message_stop`). `isStreamTruncated()` returns `false` immediately for non-sentinel transports, so a healthy Gemini response whose finishReason normalizes to `null` can never false-positive. Tests cover the non-`[DONE]` case explicitly.

**2. Socket cleanup on break — you were right to flag it: the `finally` did *not* abort.** Fixed in 7252914: the engine's `finally` now calls `controller.abort()`, so breaking the loop on `sawDone` (or any exit path) releases the connection deterministically. No-op when the body is already fully consumed.

**3. Flushes on throw — done in 3f96736.** `flushRemainingToolCalls()` + `flushReasoningFallback()` moved into a `finally` around `streamOpenCodeResponse()` in all four transports (chat-completions, responses, google, anthropic), so tool calls/reasoning already received survive a truncation error.

Full lint stack green on the branch tip.

ltmoerdani:
All three points look good now @Fahad090NP. The `finally` abort in `7252914` closes the gap I was worried about, and the flush-on-throw in `3f96736` means tool calls and reasoning survive a truncation error. I re-ran the branch locally: compile passes, all 333 unit tests green.

One small edge case left. Non-blocking for me, but easy to fix if you agree:

`normalizeResponsesFinishReason` in `src/core/routing.ts` returns `null` for `stop_reason` values it doesn't recognize (e.g. `max_tool_calls`, which is a valid OpenAI value). Since `extract.ts` only captures `finish_reason` when it's a string, an unrecognized-but-healthy `stop_reason` leaves `finishReason` undefined. If the gateway ever closes a Responses stream without `[DONE]`, that stream gets flagged as truncated even though it completed fine.

Fix would be one line: return `"stop"` as the fallback for any non-empty value instead of `null`. Happy to merge as-is either way since the gateway normally sends `[DONE]`. Your call.

Fahad090NP:
Agreed and done in the fallback commit — `normalizeResponsesFinishReason` now returns `"stop"` for any non-empty unrecognized value (e.g. OpenAI's `max_tool_calls`), so a Responses stream that closes without `[DONE]` but with a valid terminal reason is never flagged as truncated. Absent reason still maps to `null` as before. Added `src/test/routing.test.ts` covering the unrecognized-value, recognized-value, and absent-reason cases. All 7 gates green.

PR #174, opened 2026-08-21
Fix Muse Spark context window not being updated & 128k context window cap for all models

Barragek0:
## 📝 What does this change?

Fixes the Muse Spark context window staying at 0%. The Responses API nests usage under `response.usage` on the `response.completed` event, but `updateRequestUsageSummary` only checked top-level `usage`. This adds a fallback to `response.usage` and handles `input_tokens_details.cached_tokens` so prompt, completion, total, and cached tokens are reported correctly. The fix restores the native `LanguageModelDataPart` with MIME `usage` that VS Code uses for the context window indicator.

Also adds documentation that Muse Spark reasoning is currently returned as `encrypted_content`, which is not decryptable client-side, so thinking blocks remain empty until the gateway relays plaintext reasoning. See `src/thinking/muse.ts` and `src/core/routing.ts`.

## 🧪 How did you test it?

- Added 3 unit tests in `src/test/extractors.test.ts` covering nested `response.usage`, precedence over top-level usage, and the existing OpenAI shape.
- Verified with a live `muse-spark-1.2-contributor-free` trace that `response.completed` carries `input_tokens: 231407` under `response.usage`.

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works


Barragek0:
Doesn't seem to be working correctly. Context window does now surpass 0, but seems to only climb up to roughly ~13% and does not increase further, need to verify if this is a bug with the implementation.

Barragek0:
Related to #169. 512KB cap is fine for smaller context windows but doesn't scale up with context window size, meaning the context window cap for all models is incorrectly being limited to ~128k tokens. Easiest solution is to adjust the cap dependent on context window limit for the model.

Barragek0:
Added `HISTORY_BYTES_PER_TOKEN = 4.5` to `src/config.ts` and extracted the scaling logic into a new pure function `historyByteCapForBudget(inputBudgetTokens)` in `src/provider/historyTrim.ts`. The provider now calls this instead of using the hardcoded `MAX_REQUEST_PAYLOAD_BYTES`, so the byte cap dynamically scales with the token budget: `max(512KB, inputBudget × 4.5)`. 

The 512KB floor is preserved for smaller context windows (where it correctly prevents 503s), while 1M windows get a ~3.3MB byte cap - letting the 70% token budget be the actual limiter instead of the byte cap capping everything at ~128K tokens. 

Also added unit tests in `src/test/messages.test.ts` verifying the floor, linear scaling, and that the byte cap stays looser than the token budget across all window sizes (128K through 1M).

ltmoerdani:
hey, thanks for this! reviewed the whole thing and it looks good. the nested `response.usage` fallback matches the Responses API spec, and the scaled byte cap is the right call (the 512KB hard cap was silently clamping 1M-context models to ~13% of their window). tests are solid too.

one small thing: main got a new CHANGELOG entry since you branched, so there's a conflict in `CHANGELOG.md`. it's docs-only, no code conflicts, so no need to rebase on your side. i'll resolve it locally (keeping both entries) and merge with a merge commit to preserve your commits.

merging now, thanks again!

PR #175, opened 2026-08-21
fix: exclude image data from history-trim byte ceiling — fixes #173

### LINKED ISSUE #173: Vision payloads above 512KB trigger futile history trim (MAX_REQUEST_PAYLOAD_BYTES vs image data)
## Context

PR #169 added `trimOldMessagesToFitContext` with a dual budget: token cap (`HISTORY_TRIM_TARGET_RATIO = 0.7`) and a hard byte ceiling (`MAX_REQUEST_PAYLOAD_BYTES = 512 KB`, `src/config.ts`).

## Problem

Image data is counted toward the byte ceiling:

- `MAX_HISTORY_IMAGES_KEPT = 2` images survive history trimming
- Images normalize up to `MAX_IMAGE_BASE64_BYTES = 5 MB` base64 each

So a vision payload with one or two images can sit **above 512KB even after token trimming already succeeded**. The trimmer then drops the entire middle conversation history trying to get under the byte cap, and still fails (falls back to `noTrim`). Result: the user loses context for no benefit.

Also worth noting: the byte cap itself may be overly conservative. Issue #104 was closed withdrawn after the gateway was confirmed to accept bodies up to ~1.5MB (see #44), so there is no measured hard limit at 512KB.

## Suggested fix

Exclude image data (base64 `image_url` parts) from the byte accounting in `payloadBytes()` (`src/provider/historyTrim.ts`), or apply a higher byte cap for payloads carrying image parts.

Related: #109 (the real unbounded-history failure this trim solves), #78 (fetch resilience, also from #169).

Fahad090NP:
## Summary

Fixes #173 — and a real-world 503 observed on `payloadBytes=520805`: a vision payload carrying the images that survived image trimming sat above the 512 KB `MAX_REQUEST_PAYLOAD_BYTES` ceiling no matter what, so `trimOldMessagesToFitContext` dropped the whole text middle-history for nothing and still reported over-cap.

## Root cause

`payloadBytes()` stringified the raw message array; base64 `image_url.url` data dominates that count but is already bounded by the dedicated image-history trimmer (`MAX_HISTORY_IMAGES_KEPT = 2`). The byte ceiling exists to bound *text* growth, so counting image weight double-punishes vision sessions.

## Changes

- **`src/provider/historyTrim.ts`** — new `stripImageData()`: copy-on-write clone of a message with oversized base64 data URLs replaced by an `[image]` placeholder. `payloadBytes()` and per-unit byte measurement run on stripped messages; structure overhead stays counted, so the estimate remains slightly pessimistic (never under-trims). Text-only payloads measure exactly as before.
- **`src/test/messages.test.ts`** — three new cases: a >512 KB image payload triggers zero text trims; genuinely oversized *text* history still trims with images present; small non-data image URLs are counted fully.
- **Docs** — `docs/issues/75-20260821-issue173-vision-byte-cap.md` + CHANGELOG Fixed entry.

## Verification

- `npm run lint` — all 7 gates green (editorconfig, eslint, markdown, prettier, shellcheck, typecheck, unit tests).
- Staged-lint pre-commit hook passed on every commit.

PEACE BE UPON YOU


ltmoerdani:
Finally had time to test this properly. pulled the branch, ran the suite locally, all 382 pass, and the fix matches what I sketched in #173 down to the copy-on-write clone. good pick on that one.

one thing caught my eye in `stripImageData()` before merge:

```ts
if (typeof url === "string" && url.length > IMAGE_DATA_PLACEHOLDER.length) {
```

take a hosted image like `https://cdn.example.com/screenshot.png`. 42 chars, so the length check strips it down to `[image]` the same as a base64 blob. but the JSDoc and the PR description both say base64 data URLs, and a plain URL never carries the payload weight that caused #173. nothing breaks in practice since hosted URLs are tiny, I just want the code to match what the contract says. `url.startsWith("data:")` is all it takes.

the third test hints at the same gap without meaning to: it feeds a 27-char https URL and asserts `finalBytes > 0`, which holds whether the URL was stripped or counted. once the `data:` check lands, asserting the URL's full length shows up in `finalBytes` would make that test actually prove its name.

push that and I'll merge right after.

Fahad090NP:
Both points addressed and pushed:

1. `stripImageData()` now gates on `url.startsWith("data:")` — hosted URLs are never stripped, so the code matches the JSDoc/PR contract exactly.
2. The third test now proves its name: it asserts the hosted URL's full serialized length shows up in `finalBytes` (not collapsed to `[image]`), plus a new case proving every `data:` URL is stripped regardless of size.

Also rebased onto current `main` (kept both CHANGELOG entries and both test suites — the `historyByteCapForBudget` suite from #174 and the image-exclusion suite). Full gate green: typecheck + all 7 lint checks + 398/398 tests.

PR #185, opened 2026-08-22
fix(provider): cross-check models.dev deprecated status against gateway to prevent hiding live models

### LINKED ISSUE #182: [BUG] Deepseek does not show free models.
The free version of `DeepSeek v4 Flash Free` of OpenCodeZen was previously available in the Copilot list with this extension, but it has been unavailable for the past few days. Reinstalling it did not resolve the issue and I have also cleared the extension's cache.

Barragek0:
## 📝 What does this change?

Closes #182

`models.dev` `status: deprecated` was hiding live models from the picker. `laguna-s-2.1-free` is live and working on the gateway but `models.dev` incorrectly marks it deprecated. The extension's `shouldHideDeprecatedModel` filter was hiding it unconditionally.

The fix cross-checks `models.dev` deprecated status against the live gateway response. `shouldHideDeprecatedModel` now only hides when **both** agree: `models.dev` says deprecated **AND** the gateway confirms the model is absent. Offline/fallback paths fail open (no live data → don't hide on stale metadata).

This preserves the original #03 protection (gateway can list broken models, `models.dev deprecated` catches them) while fixing #182 (gateway lists working models, stale `models.dev` falsely hides them).

**Note on `deepseek-v4-flash-free`:** The reporter's original example is actually an upstream problem — the gateway lists the model but requests fail with "Model is unavailable". Neither `models.dev` nor the gateway is a reliable source of truth for availability. Our fix correctly shows the model (gateway lists it) and surfaces the error clearly when it fails. The real fix for `deepseek` is upstream.

### Files changed

| File | Change |
|---|---|
| `src/provider/settings.ts` | `shouldHideDeprecatedModel` gains `liveModelIds?: ReadonlySet<string>` parameter; returns `false` when live set confirms the model is present or when no live data is available |
| `src/provider/modelList.ts` | `filterAvailableModels` signature gains `liveModelIds?`; gateway fetch builds `Set(ids)` and passes it through |
| `src/provider/OpenCodeProvider.ts` | `filterAvailableModels` threads `liveModelIds` to `shouldHideDeprecatedModel`; fetcher wiring updated |
| `src/test/deprecatedFilter.test.ts` | 8 unit tests covering all branches (non-Zen vendor, not deprecated, absent status, offline fail-open, live gateway present, live gateway absent, agent-variant vendor, empty live set) |
| `docs/issues/78-20260822-issue182-deprecated-model-gateway-crosscheck.md` | Full issue documentation with root cause, solution, verification matrix |
| `CHANGELOG.md` | Added fix entry under `[Unreleased]` |
| `ARCHITECTURE-MAP.md` | Updated model discovery step 3 description |

### Verification

| Model | Gateway | `models.dev` | Actually works? | Before fix | After fix |
|---|---|---|---|---|---|
| `laguna-s-2.1-free` | ✅ listed | deprecated | ✅ Yes | ❌ hidden | ✅ shown |
| `deepseek-v4-flash-free` | ✅ listed | deprecated | ❌ No (400) | ❌ hidden | ✅ shown* |
| `ring-2.6-1t-free` | ❌ absent | deprecated | ❌ No | ✅ hidden | ✅ hidden |
| `trinity-large-preview-free` | ❌ absent | deprecated | ❌ No | ✅ hidden | ✅ hidden |

\* Shown but fails at runtime — upstream issue, not solvable from extension side.

## 🧪 How did you test it?

- `npm run compile` ✅
- `npm run lint` (6 checks) ✅
- `npm test` (305+ unit tests including 8 new `deprecatedFilter` tests) ✅
- `npm run package` ✅ (VSIX built)
- Live gateway check: `laguna-s-2.1-free` confirmed working ✅
- `deepseek-v4-flash-free` confirmed broken upstream (400 error) ✅

## ✅ Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] I tested it works

### DIFF
--- a/ARCHITECTURE-MAP.md
+++ b/ARCHITECTURE-MAP.md
@@ -273,7 +273,7 @@ flowchart LR
 
 1. **Key resolution**: BYOK `options.configuration.apiKey` → (if BYOK group observed, return `[]` to avoid duplicates, issue #106/#131) → `SecretStorage` fallback.
 2. Persist key to SecretStorage (non-agent variants) so agent variants inherit it.
-3. `fetchModels()` — live GET `modelsUrl` with retry/backoff/timeout (issue #78) → `filterAvailableModels()` (drops `KNOWN_UNAVAILABLE_MODEL_IDS`, deprecated Zen models, `freeOnly` filter).
+3. `fetchModels()` — live GET `modelsUrl` with retry/backoff/timeout (issue #78) → `filterAvailableModels()` (drops `KNOWN_UNAVAILABLE_MODEL_IDS`, deprecated Zen models cross-checked against gateway response (issue #182), `freeOnly` filter).
 4. Per model: `resolveModelMetadata()` → `resolveModelRouting()` → `modelLimits()` → `modelCapabilities()` → `modelConfigurationSchema()` (thinking submenu + context-size tier) → build `OpenCodeModel` (general variant or `::agent-host` variant with `targetChatSessionType: "copilotcli"`).
 
 ### 5.3 Chat Request (`provideLanguageModelChatResponse`)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to the **OpenCode Go BYOK Provider** extension are documente
 
 ### Fixed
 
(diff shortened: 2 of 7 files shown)


Barragek0:
I've left a comment in the base issue that this PR closes explaining the issue further and why I think this PR is the right approach.

ltmoerdani:
Clean fix. i went through the diff and the cross-check approach makes sense. keeping the #03 protection while un-hiding models the gateway still serves is the right call, and failing open when there's no live data is what i'd want too.

good catch on `deepseek-v4-flash-free` being upstream. saves us from chasing a bug we can't fix on our side.

one small thing, not a blocker. with fail-open, models that were previously hidden via stale `deprecated` flags will now show up in the fallback/offline path too. `fallbackModels` is a curated list so i think the impact is minimal, but wanted to note the trade-off in case someone hits it later.

tests cover all the branches, docs and changelog are in place, CI green. merging this as-is.

Issue #171, opened 2026-08-21
[BUG] deepseek-v4-flash sends max_tokens=384000 but OpenCode Go now caps completion at 131072


Skywind5487:
## 🐛 What happened?

Using `deepseek-v4-flash` through OpenCode Go on extension `0.6.0`, the extension sends `max_tokens: 384000` and the provider rejects the request immediately with HTTP 400:

```text
bad request: max_tokens is too large: 384000. This model supports at most 131072 completion tokens.
```

This appears to be a model-metadata / upstream-limit mismatch. The request payload itself is not especially large (`payloadBytes=184129`), so this is different from a total-context overflow.

There is a related older issue, #109, where `384000` caused `prompt + completion` to exceed the context window. This report is slightly different: the current OpenCode Go upstream explicitly rejects `384000` because the model's completion-token limit is now `131072`.

## 🔄 Steps to reproduce

1. Install extension version `0.6.0`.
2. Configure OpenCode Go.
3. Select `deepseek-v4-flash`.
4. Send a request in GitHub Copilot Chat.
5. Observe that the extension sends `max_tokens=384000`.
6. OpenCode Go returns HTTP 400 because the upstream accepts at most `131072` completion tokens.

## ✅ What did you expect?

The extension should not send a `max_tokens` value above the limit currently accepted by OpenCode Go for the selected model.

Ideally, the effective output limit should be capped to the provider/model limit (currently reported as `131072`), or otherwise avoid trusting stale metadata when the upstream limit has changed.

## 🖥️ Environment

- **Extension version:** `0.6.0`
- **VS Code version:** not captured
- **OS:** Windows
- **Provider:** OpenCode Go
- **Model:** `deepseek-v4-flash`

## 📸 Logs or screenshots (optional)

```text
Sorry, your request failed. Please try again.

Client Request Id: 246eb16c-60cb-411c-bb94-4d7a82540b68

Reason: OpenCode Go API request failed (400) model=deepseek-v4-flash payloadBytes=184129: Error from provider (Console Go): Upstream request failed: [bad_request] bad request: max_tokens is too large: 384000. This model supports at most 131072 completion tokens.: OpenCodeRequestError: OpenCode Go API request failed (400) model=deepseek-v4-flash payloadBytes=184129: Error from provider (Console Go): Upstream request failed: [bad_request] bad request: max_tokens is too large: 384000. This model supports at most 131072 completion tokens.
[...]

PR #191, opened 2026-08-24
fix(retry): degrade optional params on [1210] invalid input 400s (#190)

### LINKED ISSUE #190: [BUG] OpenCode

Client Request Id: 9040dc4d-6526-42aa-8144-ec792daceaaf

Reason: OpenCode Go API request failed (400) model=ox-alpha-free payloadBytes=1983985: Error from provider (Console Go): Upstream request failed: [1210] Invalid API parameter, please check the documentation.invalid input: OpenCodeRequestError: OpenCode Go API request failed (400) model=ox-alpha-free payloadBytes=1983985: Error from provider (Console Go): Upstream request failed: [1210] Invalid API parameter, please check the documentation.invalid input at buildOpenCodeRequestError (/home/fahad090np/.vscode/extensions/ltmoerdani.opencode-copilot-chat-0.7.0/out/errors.js:39:12) at streamOpenCodeResponse (/home/fahad090np/.vscode/extensions/ltmoerdani.opencode-copilot-chat-0.7.0/out/transports/engine.js:278:73) at process.processTicksAndRejections (node:internal/process/task_queues:104:5) at async streamChatCompletions (/home/fahad090np/.vscode/extensions/ltmoerdani.opencode-copilot-chat-0.7.0/out/transports/chatCompletions.js:26:9) at async OpenCodeProvider.provideLanguageModelChatResponse (/home/fahad090np/.vscode/extensions/ltmoerdani.opencode-copilot-chat-0.7.0/out/provider/OpenCodeProvider.js:382:13)

Note: GitHub is currently experiencing a service disruption. This may be affecting Copilot. Check GitHub Status for details.

Fahad090NP:
Fixes #190

## Summary

`ox-alpha-free` rejects requests with HTTP 400:

```text
Upstream request failed: [1210] Invalid API parameter, please check the documentation.invalid input
```

The gateway gives **no hint which parameter** is invalid, and the reporter's payload was 1.98 MB — well-formed otherwise. Nothing in our request builder is obviously wrong for this model (it's a fallback-family model, so no thinking fields are even sent), which means one of the optional parameters we always include isn't accepted by this upstream.

## Fix

New `patchInvalidInput` classifier in `src/retry.ts`: when a 400 body matches `[1210] … invalid input / invalid API parameter`, the retry degrades the request by stripping optional parameters **one per attempt**, least-likely-to-matter first:

1. `stream_options` (usage accounting hint — pure nicety)
2. `temperature` (fall back to the model's own default)
3. image parts (some models reject data-URL images despite metadata claiming vision support; text parts are preserved)

Each patch only fires when the parameter is actually present, so once everything optional is gone the sequence stops and real failures surface normally — no infinite retries, and no masking of genuine errors after degradation is exhausted.

This mirrors how #171 (completion-cap) and #109 (context overflow) self-heal: use the engine's existing 400-retry to adapt to upstream constraints we can't know ahead of time.

## Verification

- `npm run lint` all 7 checks green.
- 5 new unit tests in `src/test/retry.test.ts`: each degradation step, image-part stripping while preserving text, exhaustion → undefined, and no-fire for unrelated 400s.

Fi Amanillah


ltmoerdani:
Nice one @Fahad090NP, the classifier and tests look clean. one thing before I merge though.

the ladder doesn't actually run end-to-end right now. `analyzeHttp400ForRetry` returns one patch per call, but `src/transports/engine.ts` only calls it once and retries once. so if dropping `stream_options` still gets a 400, we stop there. `temperature` and the image strip are never reached in a single request, the unit tests pass because they call the analyzer directly with pre-degraded bodies, not through the engine.

the JSDoc says "one per engine retry attempt", which is why I'm asking for the fix rather than a doc edit. could you wrap the analyze → patch → retry in `engine.ts` in a loop? something like:

```ts
let response = await fetchWithTransientRetry(payload);
let consumedErrorBody: string | undefined;
for (let patchAttempt = 0; patchAttempt < 3 && response.status === 400; patchAttempt++) {
  const errorDetail = consumedErrorBody ?? (await response.text());
  const parsedBody = JSON.parse(rawPayload) as Record<string, unknown>;
  const patch = analyzeHttp400ForRetry(errorDetail, parsedBody);
  if (!patch) break;
  options.output?.appendLine(`[retry] HTTP 400 recoverable: ${patch.reason}. Retrying with patched body…`);
  payload = JSON.stringify(patch.body);
  rawPayload = payload; // keep them in sync for the next iteration
  response = await fetchWithTransientRetry(payload);
  consumedErrorBody = response.status === 400 ? await response.text() : undefined;
}
```

(key detail: re-analyze has to run against the patched body, so `rawPayload` needs to be updated inside the loop too.)

a couple of notes while you're in there:

- the `[http-error-body]` log line from the current single-attempt version should stay, just move it inside the loop or before it.
- no need to add engine-level tests if it's a pain, the existing analyzer tests plus a compile pass are fine. if you do want one, a loop test with a mock fetch that 400s until images are stripped would be the one worth having.

happy to merge once the loop is in. thanks for picking this up.

Fahad090NP:
Loop is in. `engine.ts` now wraps the analyze → patch → retry in a `for` loop up to `MAX_400_PATCH_ATTEMPTS = 3`:

- each iteration re-analyzes the latest 400 body against the **latest patched payload** — `rawPayload` is reassigned inside the loop as you flagged, so the next ladder step (temperature → images) is visible to the analyzer
- `consumedErrorBody` carries between iterations: the retry's 400 body is consumed once and reused for the next analysis / final error message, no double-read
- `[http-error-body]` log moved inside the loop
- exits when status ≠ 400 or the analyzer returns no patch

Also rebased onto current `main` (picked up #188's merge). Skipped the engine-level mock-fetch test per your note — the analyzer tests cover the ladder steps. Full gate green (all 7 checks). Fi Amanillah

ltmoerdani:
Nice @Fahad090NP, the loop looks right, `rawPayload` sync, `consumedErrorBody` handling, `break` on no patch, all clean. CI green too.

merging now.

